PR #25, opened 2024-08-18
Update device provides

danielskinstad:


core/src/mender-utils.c
```diff
@@ -509,3 +509,161 @@ mender_utils_append_list(mender_key_value_list_t **list1, mender_key_value_list_
...
+    }
+
+    free(to_free->key);
+    free(to_free->value);
+    free(to_free);
+    return MENDER_OK;
+}
+
+bool
+mender_utils_compare_wildcard(char *string, char *wildcard_string) {
```
larsewi, line 567: The variable name `string` might conflict with the type `string` if we ever want to compile for C++. You can consider `s/string/str/`
```suggestion
mender_utils_compare_wildcard(const char *string, const char *wildcard_string) {
```

core/src/mender-utils.c
```diff
@@ -509,3 +509,161 @@ mender_utils_append_list(mender_key_value_list_t **list1, mender_key_value_list_
...
+
+    char *escaped = malloc(strlen(wildcard_string) + 1);
+    if (NULL == escaped) {
+        mender_log_error("Unable to allocate memory for escaped string");
+        ret = false;
+        goto END;
+    }
+
+    /* Split string into substrings on each wildcard */
+    for (size_t i = 0; i < strlen(wildcard_string); i++) {
```
larsewi, line 586: Consider calling `strlen(wildcard_string)` once - before the `malloc` above.  Assign it to a constant variable, and use that instead. This way you will save a few CPU cycles.

core/src/mender-utils.c
```diff
@@ -509,3 +509,161 @@ mender_utils_append_list(mender_key_value_list_t **list1, mender_key_value_list_
...
+    char **substrings     = NULL;
+    size_t escaped_pos    = 0;
+    bool   found_wildcard = false;
+    bool   ret            = true;
+
+    char *escaped = malloc(strlen(wildcard_string) + 1);
+    if (NULL == escaped) {
+        mender_log_error("Unable to allocate memory for escaped string");
+        ret = false;
+        goto END;
```
larsewi, line 582: Nothing to clean up yet.
```suggestion
        return false;
```

core/src/mender-utils.c
```diff
@@ -509,3 +509,161 @@ mender_utils_append_list(mender_key_value_list_t **list1, mender_key_value_list_
...
+    }
+
+    /* Split string into substrings on each wildcard */
+    for (size_t i = 0; i < strlen(wildcard_string); i++) {
+        if ('*' == wildcard_string[i]) {
+            found_wildcard       = true;
+            escaped[escaped_pos] = '\0';
+
+            substrings = realloc(substrings, (substrings_pos + 1) * sizeof(char *));
```
larsewi, line 591: > For realloc(), the input pointer is still valid if reallocation failed.
> - realloc(3)

If `realloc` fails here, it would return `NULL`, and you would lose your only reference to the original "`substrings`". Hence, this memory would be leaked. Consider assigning the return value to a temporary variable and check it for `NULL` before assigning that value to `substrings`.

core/src/mender-utils.c
```diff
@@ -509,3 +509,161 @@ mender_utils_append_list(mender_key_value_list_t **list1, mender_key_value_list_
...
+
+            substrings_pos++;
+            escaped_pos = 0;
+        } else {
+            escaped[escaped_pos++] = wildcard_string[i];
+        }
+    }
+
+    escaped[escaped_pos] = '\0';
+    substrings           = realloc(substrings, (substrings_pos + 1) * sizeof(char *));
```
larsewi, line 613: Same here :)

core/src/mender-utils.c
```diff
@@ -509,3 +509,161 @@ mender_utils_append_list(mender_key_value_list_t **list1, mender_key_value_list_
...
+        item = item->next;
+    }
+
+    free(to_free->key);
+    free(to_free->value);
+    free(to_free);
+    return MENDER_OK;
+}
+
+bool
```
larsewi, line 566: This function should probably have the ability to return error if any of the memory allocations fail?

core/src/mender-utils.c
```diff
@@ -509,3 +509,161 @@ mender_utils_append_list(mender_key_value_list_t **list1, mender_key_value_list_
...
+            }
+        }
+    }
+
+    *list2 = NULL;
+    return MENDER_OK;
+}
+
+mender_err_t
+mender_utils_key_value_list_delete_node(mender_key_value_list_t **list, char *key) {
```
larsewi, line 540:
```suggestion
mender_utils_key_value_list_delete_node(mender_key_value_list_t **list, const char *key) {
```

platform/storage/posix/src/mender-storage.c
```diff
@@ -228,8 +228,8 @@ mender_storage_delete_device_config(void) {
 #ifdef CONFIG_MENDER_FULL_PARSE_ARTIFACT
 #ifdef CONFIG_MENDER_PROVIDES_DEPENDS
 
-mender_err_t
-mender_storage_set_provides(mender_key_value_list_t *provides) {
+static mender_err_t
+mender_set_provides(mender_key_value_list_t *provides, char *filename) {
```
larsewi, line 232:
```suggestion
mender_set_provides(mender_key_value_list_t *provides, const char *filename) {
```

platform/storage/zephyr/nvs/src/mender-storage.c
```diff
@@ -290,14 +291,24 @@ mender_storage_set_provides(mender_key_value_list_t *provides) {
...
+    return mender_set_provides(provides, MENDER_STORAGE_NVS_PROVIDES);
+}
+
+mender_err_t
+mender_storage_set_new_provides(mender_key_value_list_t *new_provides) {
+    return mender_set_provides(new_provides, MENDER_STORAGE_NVS_NEW_PROVIDES);
+}
+
+static mender_err_t
+mender_get_provides(mender_key_value_list_t **provides, char *filename) {
```
larsewi, line 304:
```suggestion
mender_get_provides(mender_key_value_list_t **provides, const char *filename) {
```

larsewi (changes requested):
Good job 🚀 Some nitpicks only.

core/src/mender-utils.c
danielskinstad: Yeah, I agree

platform/storage/zephyr/nvs/src/mender-storage.c
```diff
@@ -279,7 +280,7 @@ mender_storage_set_provides(mender_key_value_list_t *provides) {
     }
 
     /* Write provides */
-    if (nvs_write(&mender_storage_nvs_handle, MENDER_STORAGE_NVS_PROVIDES, provides_str, strlen(provides_str) + 1) < 0) {
+    if (nvs_write(&mender_storage_nvs_handle, filename, provides_str, strlen(provides_str) + 1) < 0) {
```
vpodzime, line 283: `nvs_write()` takes a handle and then an integer, right? `filename` is a `char *`.

platform/storage/posix/src/mender-storage.c
```diff
@@ -248,13 +248,23 @@ mender_storage_set_provides(mender_key_value_list_t *provides) {
...
+    return mender_set_provides(provides, MENDER_STORAGE_NVS_PROVIDES);
+}
+
+mender_err_t
+mender_storage_set_new_provides(mender_key_value_list_t *new_provides) {
+    return mender_set_provides(new_provides, MENDER_STORAGE_NVS_NEW_PROVIDES);
+}
+
+static mender_err_t
+mender_get_provides(mender_key_value_list_t **provides, char *filename) {
```
vpodzime, line 261: `const char *filename`.

core/src/mender-utils.c
```diff
@@ -509,3 +509,161 @@ mender_utils_append_list(mender_key_value_list_t **list1, mender_key_value_list_
     *list2 = NULL;
     return MENDER_OK;
 }
+
+mender_err_t
+mender_utils_combine_unique(mender_key_value_list_t **list1, mender_key_value_list_t **list2) {
```
vpodzime, line 514: Below it says the list is appended, why not call the function `append_unique()` then?

core/src/mender-utils.c
```diff
@@ -509,3 +509,161 @@ mender_utils_append_list(mender_key_value_list_t **list1, mender_key_value_list_
...
+    for (mender_key_value_list_t *item = *list2; item; item = item->next) {
+        bool key_exists = false;
+        for (mender_key_value_list_t *item1 = *list1; item1; item1 = item1->next) {
+            if (0 == strcmp(item->key, item1->key)) {
+                key_exists = true;
+                break;
+            }
+        }
+        if (!key_exists) {
+            if (MENDER_OK != mender_utils_create_key_value_node(item->key, item->value, list1)) {
```
vpodzime, line 528: We should not duplicate the `key` and `value` here, let's just re-use them.

core/src/mender-utils.c
```diff
@@ -509,3 +509,161 @@ mender_utils_append_list(mender_key_value_list_t **list1, mender_key_value_list_
...
+                mender_log_error("Unable to create key-value node");
+                return MENDER_FAIL;
+            }
+        }
+    }
+
+    *list2 = NULL;
```
vpodzime, line 535: This leaks the whole `list2` AFAICT.

include/mender-utils.h
```diff
@@ -210,6 +210,22 @@ mender_err_t mender_utils_key_value_list_to_string(mender_key_value_list_t *list
...
+/**
+ * @brief Delete a node from a linked list
+ */
+mender_err_t mender_utils_key_value_list_delete_node(mender_key_value_list_t **list, char *key);
+
+/**
+ * @brief Compare `string` with `wild_card_string`
+ * @return true if matches, else false
+ */
+bool mender_utils_compare_wildcard(char *string, char *wildcard_string);
```
vpodzime, line 227: `const char *`

core/src/mender-utils.c
vpodzime, line 567: What do we need wildcard comparison for? `clears_provides` uses regular expressions, IIRC. Is it for something else?

core/src/mender-client.c
```diff
@@ -838,7 +850,24 @@ mender_client_authentication_work_function(void) {
...
+            if (MENDER_OK != mender_storage_get_new_provides(&new_provides)) {
+                mender_log_error("Unable to get new provides");
+                mender_client_publish_deployment_status(id, MENDER_DEPLOYMENT_STATUS_FAILURE);
+                goto RELEASE;
+            }
+            if (MENDER_OK != mender_storage_set_provides(new_provides)) {
+                mender_log_error("Unable to set provides");
+                mender_client_publish_deployment_status(id, MENDER_DEPLOYMENT_STATUS_FAILURE);
+                goto RELEASE;
+            }
```
vpodzime, line 866: This block is worth a comment, it's quite confusing what's going on in here. Also, do we really want to do this right after authentication?

vpodzime (commented):
My 2 cents...

core/src/mender-utils.c
danielskinstad: It's based on this I wrote in mender: https://github.com/mendersoftware/mender/blob/eae7540db3ba5c7fea256f217dab956a5e06bb30/src/mender-update/context/context.cpp#L236 - but this was written to avoid c++ exceptions which is not a problem here.

It only matches [wildcards](https://github.com/mendersoftware/mender-artifact/blob/master/Documentation/artifact-format-v3.md#clears_artifact_provides). I saw a thread where Joel asked about regex for zephyr, and he solved it using [picolibc](https://github.com/zephyrproject-rtos/zephyr/issues/53407), should we do something like that? I think we only  need to match wildcards.

core/src/mender-client.c
danielskinstad: This will happen after a reboot and if the deployment is a success, right?. I didn't find anywhere else to put this, and figured this was the closest I came to `ArtifactCommit` after a reboot. Do you have any other suggestions?

core/src/mender-utils.c
vpodzime: Oh! I see! Looks like `*` is the only thing we need to care about. [1] I didn't know that. Examples on the Internet (incl. our own docs) suggest it's a regex, but that's a different story. Let's not worry about that now. :+1: 

[1] https://github.com/mendersoftware/mender-artifact/blob/master/Documentation/artifact-format-v3.md#clears_artifact_provides

core/src/mender-client.c
vpodzime: I guess there's no better place until we have the update module state machine and it can go into the `Commit` state. But the comment is necessary.

core/src/mender-utils.c
```diff
@@ -509,3 +509,200 @@ mender_utils_append_list(mender_key_value_list_t **list1, mender_key_value_list_
...
+    substrs_pos++;
+
+    ret = MENDER_OK;
+
+    /* If no wildcard was found, perform a direct str comparison */
+    if (!found_wildcard) {
+        if (0 != strcmp(str, wildcard_str)) {
+            *match = false;
+            goto END;
+        }
```
larsewi, line 673: What if no wildcard was found, but the strings are equal?
```suggestion
        *match = (0 == strcmp(str, wildcard_str));
        goto END;
```

larsewi (commented):
Just a few comments. Otherwise it looks good 🚀

core/src/mender-utils.c
danielskinstad: Good catch! I forgot to re-base some changes to my unit-test branch so I didn't catch this

larsewi (approved):
Great work 🚀

platform/storage/posix/src/mender-storage.c
```diff
@@ -248,13 +248,23 @@ mender_storage_set_provides(mender_key_value_list_t *provides) {
...
+
+static mender_err_t
+mender_get_provides(mender_key_value_list_t **provides, const char *filename) {
 
     assert(NULL != provides);
 
     char  *provides_str = NULL;
     size_t provides_length;
-    if (MENDER_OK != mender_storage_read_file(MENDER_STORAGE_NVS_PROVIDES, (void **)&provides_str, &provides_length)) {
+    if (MENDER_OK != mender_storage_read_file(filename, (void **)&provides_str, &provides_length)) {
```
vpodzime, line 267: A bit out of scope of this PR, but I don't see why this code uses the generic `mender_storage_read_file()` when it is a POSIX-specific code that can use the POSIX-specific code to read the file contents.

platform/storage/zephyr/nvs/src/mender-storage.c
```diff
@@ -268,8 +267,10 @@ mender_storage_delete_device_config(void) {
     return MENDER_OK;
 }
 
-mender_err_t
-mender_storage_set_provides(mender_key_value_list_t *provides) {
+#ifdef CONFIG_MENDER_FULL_PARSE_ARTIFACT
+#ifdef CONFIG_MENDER_PROVIDES_DEPENDS
+static mender_err_t
+mender_set_provides(mender_key_value_list_t *provides, const int filename) {
```
vpodzime, line 273: A bit strange to call the argument `filename` is this Zephyr-specific code where it is not a file name.

platform/storage/posix/src/mender-storage.c
```diff
@@ -248,13 +248,23 @@ mender_storage_set_provides(mender_key_value_list_t *provides) {
...
+    return mender_set_provides(provides, MENDER_STORAGE_NVS_PROVIDES);
+}
+
+mender_err_t
+mender_storage_set_new_provides(mender_key_value_list_t *new_provides) {
+    return mender_set_provides(new_provides, MENDER_STORAGE_NVS_NEW_PROVIDES);
+}
+
+static mender_err_t
+mender_get_provides(mender_key_value_list_t **provides, const char *filename) {
```
vpodzime, line 261: The difference in name between `mender_storage_get_provides()` and `mender_get_provides()` here can be quite confusing. The `mender_` prefix is not needed here for a `static` function, but maybe the name could reflect it reads provides from a file?

platform/storage/zephyr/nvs/src/mender-storage.c
```diff
@@ -290,14 +291,24 @@ mender_storage_set_provides(mender_key_value_list_t *provides) {
...
+    return mender_set_provides(provides, MENDER_STORAGE_NVS_PROVIDES);
+}
+
+mender_err_t
+mender_storage_set_new_provides(mender_key_value_list_t *new_provides) {
+    return mender_set_provides(new_provides, MENDER_STORAGE_NVS_NEW_PROVIDES);
+}
+
+static mender_err_t
+mender_get_provides(mender_key_value_list_t **provides, const int filename) {
```
vpodzime, line 304: Same here with the function name.

core/src/mender-utils.c
```diff
@@ -509,3 +509,198 @@ mender_utils_append_list(mender_key_value_list_t **list1, mender_key_value_list_
     *list2 = NULL;
     return MENDER_OK;
 }
+
+mender_err_t
+mender_utils_append_unique(mender_key_value_list_t **list1, mender_key_value_list_t **list2) {
```
vpodzime, line 514: `key_value_list_append`, it's a list-specific function.

core/src/mender-utils.c
```diff
@@ -509,3 +509,198 @@ mender_utils_append_list(mender_key_value_list_t **list1, mender_key_value_list_
...
+    size_t       substrs_pos    = 0;
+    char       **substrs        = NULL;
+    char       **tmp_substrs    = NULL;
+    size_t       escaped_pos    = 0;
+    bool         found_wildcard = false;
+    mender_err_t ret            = MENDER_FAIL;
+
+    *match = true;
+
+    char *escaped = malloc(strlen(wildcard_str) + 1);
```
vpodzime, line 612: Why `escaped`? It's a copy of the string with `'*'` replaced by `'\0'` and thus split into substrings, right?

core/src/mender-utils.c
```diff
@@ -509,3 +509,198 @@ mender_utils_append_list(mender_key_value_list_t **list1, mender_key_value_list_
...
+                *match = false;
+                goto END;
+            }
+
+            substrs_pos++;
+            escaped_pos = 0;
+        } else {
+            escaped[escaped_pos++] = wildcard_str[i];
+        }
+    }
```
vpodzime, line 648: AFAIK, we don't need `escaped` here at all or we don't need `substrs`. Either create a copy of `wildcard_str` and replace the wildcards with zeroes and thus split the string in-place. Or use `strchr(wildcard_str, '*')` in combination with `strndup()` to copy the substrings from `wildcard_str` into `substrs`.

core/src/mender-client.c
```diff
@@ -926,51 +974,72 @@ mender_compare_device_types(const char  *device_type_artifact,
...
-    assert(NULL != mender_artifact_ctx);
     mender_err_t ret = MENDER_FAIL;
+    /* Clears provides */
+    bool matches;
+    for (size_t i = 0; i < mender_artifact_ctx->payloads.size; i++) {
+        for (size_t j = 0; j < mender_artifact_ctx->payloads.values[i].clears_provides_size; j++) {
+            const char *to_clear = mender_artifact_ctx->payloads.values[i].clears_provides[j];
+            for (mender_key_value_list_t *item = *stored_provides; NULL != item; item = item->next) {
+                if (MENDER_OK != mender_utils_compare_wildcard(item->key, to_clear, &matches)) {
+                    mender_log_error("Unable to compare wildcard");
```
vpodzime, line 990: The error message should mention what kind of wildcard it was and what it was doing. Otherwise it's quite useless for the user.

core/src/mender-client.c
```diff
@@ -926,51 +974,72 @@ mender_compare_device_types(const char  *device_type_artifact,
...
+                    goto END;
+                }
+                if (matches && MENDER_OK != mender_utils_key_value_list_delete_node(stored_provides, item->key)) {
+                    mender_log_error("Unable to delete node");
```
vpodzime, line 994: Same here.

vpodzime (commented):
Looks good to me otherwise.

platform/storage/zephyr/nvs/src/mender-storage.c
```diff
@@ -39,6 +39,7 @@
 #define MENDER_STORAGE_NVS_DEPLOYMENT_DATA 3
 #define MENDER_STORAGE_NVS_DEVICE_CONFIG   4
 #define MENDER_STORAGE_NVS_PROVIDES        5
+#define MENDER_STORAGE_NVS_NEW_PROVIDES    6
```
lluiscampos, line 42: The "new provides" is really part of the Deployment Data -> info about the ongoing deployment with an uncommitted Artifact. Can't we extend the existing key for that?

danielskinstad: I could probably do that; I can add the parsed provides data as a string under a key like "new_provides" in `deployment-data.json`

lluiscampos: That would make more sense to me, and would align with the C++ client. For example if we consolidate we can clear both in the clean state, or check only one place to assume there is or there is not a deployment in progress.

danielskinstad: Yeah, I didn't think about doing it that way, but that makes a lot of sense.

core/src/mender-utils.c
```diff
@@ -509,3 +509,139 @@ mender_utils_append_list(mender_key_value_list_t **list1, mender_key_value_list_
...
+            return MENDER_OK;
+        }
+    }
+
+    /*
+     * Iterate over substrings separated by wildcard *
+     * Attempt to find the substring in the string
+     */
+    while (NULL != (ptr = strchr(boundary, '*'))) {
+        const int len = (int)(ptr - boundary);
```
vpodzime, line 629: `size_t`

core/src/mender-utils.c
```diff
@@ -509,3 +509,139 @@ mender_utils_append_list(mender_key_value_list_t **list1, mender_key_value_list_
...
+        char      substr[len + 1];
+        substr[len] = '\0';
+        strncpy(substr, boundary, len);
+        const char *find = strstr(to_match, substr);
```
vpodzime, line 633: You can use `memcmp()` and avoid the necessity to have a null-terminated string and thus avoid the copy.

vpodzime (commented):
A very nice improvement! Just one last thing.

vpodzime (approved):
Nice and with no memory allocations in the wildcard comparison! :clap:  Just please double-check it (still) works. :grin: It might be worth copying the code into a separate `.c` file, adding `main()` and testing it manually a bit.

core/src/mender-client.c
```diff
@@ -1061,18 +1137,21 @@ mender_client_update_work_function(void) {
...
+        mender_log_error("Unable to prepare new provides");
         mender_client_publish_deployment_status(deployment->id, MENDER_DEPLOYMENT_STATUS_FAILURE);
         if (mender_client_deployment_needs_set_pending_image) {
             mender_flash_abort_deployment(mender_client_flash_handle);
         }
         goto END;
     }
-#endif /* CONFIG_MENDER_PROVIDES_DEPENDS */
-#endif /* CONFIG_MENDER_FULL_PARSE_ARTIFACT */
+    cJSON_AddStringToObject(mender_client_deployment_data, "provides", new_provides);
```
lluiscampos, line 1152: :clap:

PR #26, opened 2024-08-20
feat: do not check if artifact is already installed

larsewi:
The Mender Client used to check if the downloaded artifact was already installed in the past. However, this check was removed in https://github.com/mendersoftware/mender/pull/1106/commits/1fb62236abdee6d2ed2452592d289320aea1bfa5 as it prevents the use of the "Force update" option, allowing a device to take part in the same deployment twice. Today, all the "already-installed" logic is performed in the backend.

Ticket: MEN-7428
Changelog: Title
Signed-off-by: Lars Erik Wik <lars.erik.wik@northern.tech>

### DIFF
--- a/core/src/mender-client.c
+++ b/core/src/mender-client.c
@@ -999,13 +999,6 @@ mender_client_update_work_function(void) {
         goto END;
     }
 
-    /* Check if artifact is already installed (should not occur) */
-    if (!strcmp(deployment->artifact_name, mender_client_config.artifact_name)) {
-        mender_log_error("Artifact is already installed");
-        mender_client_publish_deployment_status(deployment->id, MENDER_DEPLOYMENT_STATUS_ALREADY_INSTALLED);
-        goto END;
-    }
-
     /* Reset flags */
     mender_client_deployment_needs_set_pending_image = false;
     mender_client_deployment_needs_restart           = false;


core/src/mender-utils.c
```diff
@@ -157,8 +157,6 @@ mender_utils_deployment_status_to_string(mender_deployment_status_t deployment_s
         return "success";
     } else if (MENDER_DEPLOYMENT_STATUS_FAILURE == deployment_status) {
         return "failure";
-    } else if (MENDER_DEPLOYMENT_STATUS_ALREADY_INSTALLED == deployment_status) {
-        return "already-installed";
```
lluiscampos, line 161: Strictly speaking this (and the following change) we don't need to remove. It is a valid state, although the current state machine won't use it. But what if a custom Update Module wants to do some extra checks or something custom on a peripheral or something and then return no-op to the server? I would let it be.

If we really want to save the bytes, then compile it out with an ifdef for future use.

lluiscampos (commented):
It is not an strong opinion - you can convince me otherwise. See comment below.

core/src/mender-utils.c
larsewi: You make a valid point. My only motivation for removing it, was that the original commit in the GO client removed it. However, the state is present in the C++ client, although never used. Furthermore, the ticket says the C++ client is the source of truth. So I see to it, that this change is removed 😉

lluiscampos (approved):

PR #28, opened 2024-08-23
Device compatibility

danielskinstad:


core/src/mender-client.c
```diff
@@ -1049,6 +1055,50 @@ mender_prepare_new_provides(mender_artifact_ctx_t *mender_artifact_ctx, char **n
...
+            goto END;
+        }
+    }
+
+    /* Match depends from artifact with device's provides */
+    for (mender_key_value_list_t *depends_item = depends; NULL != depends_item; depends_item = depends_item->next) {
+        bool matches = false;
+        for (mender_key_value_list_t *provides_item = stored_provides; NULL != provides_item; provides_item = provides_item->next) {
+            /* Match key-value from depends with provides */
+            if (0 == strncmp(depends_item->key, provides_item->key, strlen(depends_item->key))
```
larsewi, line 1084: Since these are null-byte terminated string, you may as well use `strcmp()`. Using `strlen()` and `strncmp()` causes the string to be iterated twice, instead of once.

vpodzime: And I guess it's time to introduce `#define StringEqual(str1, str2)...`

core/src/mender-artifact.c
```diff
@@ -211,18 +211,18 @@ mender_artifact_process_data(mender_artifact_ctx_t *ctx,
         } else if (MENDER_ARTIFACT_STREAM_STATE_PARSING_DATA == ctx->stream_state) {
 
             /* Treatment depending of the file name */
-            if (!strcmp(ctx->file.name, "version")) {
+            if (!StringEqual(ctx->file.name, "version")) {
```
larsewi, line 214: `strcmp()` will return zero (i.e., `false`) if the strings equal and non-zero otherwise (i.e., `true`). `strcmp() == 0` will do the opposite. Hence, you will need to remove the negation (i.e., `!`) here and many other places.

larsewi (changes requested):

core/src/mender-client.c
```diff
@@ -1050,6 +1055,49 @@ mender_prepare_new_provides(mender_artifact_ctx_t *mender_artifact_ctx, char **n
...
+            goto END;
+        }
+    }
+
+    /* Match depends from artifact with device's provides */
+    for (mender_key_value_list_t *depends_item = depends; NULL != depends_item; depends_item = depends_item->next) {
+        bool matches = false;
+        for (mender_key_value_list_t *provides_item = stored_provides; NULL != provides_item; provides_item = provides_item->next) {
+            /* Match key-value from depends with provides */
+            if (StringEqual(depends_item->key, provides_item->key) && (StringEqual(depends_item->value, provides_item->value))) {
```
larsewi, line 1084: This set of parentheses do more harm than good in terms of code readability.
```suggestion
            if (StringEqual(depends_item->key, provides_item->key) && StringEqual(depends_item->value, provides_item->value)) {
```

core/src/mender-artifact.c
danielskinstad: Oops, went a bit too fast there

larsewi (approved):
Great 🚀

include/mender-utils.h
```diff
@@ -34,6 +34,12 @@ extern "C" {
 #include <assert.h>
 #include <cJSON.h>
 
+/**
+ * @brief Macro for comparing two strings
+ * @return true if the strings are equal, false otherwise
+ */
+#define StringEqual(str1, str2) (0 == strcmp(str1, str2))
```
vpodzime, line 41: We might need to extend this to gracefully handle `NULL` values in the future, but good enough for now. :+1:

core/src/mender-client.c
```diff
@@ -1049,6 +1055,49 @@ mender_prepare_new_provides(mender_artifact_ctx_t *mender_artifact_ctx, char **n
...
+        bool matches = false;
+        for (mender_key_value_list_t *provides_item = stored_provides; NULL != provides_item; provides_item = provides_item->next) {
+            /* Match key-value from depends with provides */
+            if (StringEqual(depends_item->key, provides_item->key) && StringEqual(depends_item->value, provides_item->value)) {
+                matches = true;
+                break;
+            }
+        }
+        if (!matches) {
+            mender_log_error("Missing '%s:%s' in provides, required by artifact depends", depends_item->key, depends_item->value);
```
vpodzime, line 1090: I think it would be nice to log a different error in case of a value mismatch.

core/src/mender-client.c
```diff
@@ -1138,6 +1187,16 @@ mender_client_update_work_function(void) {
...
+    /* Compare Artifact's depends with the stored provides */
+    if (MENDER_OK != mender_check_device_compatibility(mender_artifact_ctx)) {
+        /* Errors logged by function */
+        mender_client_publish_deployment_status(deployment->id, MENDER_DEPLOYMENT_STATUS_FAILURE);
+        if (mender_client_deployment_needs_set_pending_image) {
+            mender_flash_abort_deployment(mender_client_flash_handle);
+        }
+        goto END;
+    }
+
```
vpodzime, line 1199: OK for now because we are refactoring this part of the code, but as I asked for before -- please don't add more code into this function. We need to merge this with `mender_compare_device_types()` into a more comprehensive `check()` function.

vpodzime (commented):
My two cents...

include/mender-utils.h
larsewi: Maybe that should be another macro.  E.g.:
```C
#define SafeStringEqual(str1, str2) ...
```
to avoid adding more code in places where it is not needed?

core/src/mender-client.c
larsewi: What error message would you like to see?

vpodzime: Well a missing key-value pair for a given key is a different thing than a value mismatch for the given key, right?

include/mender-utils.h
vpodzime: Then I vote for `StringEqual()` and `UnsafeStringEqual()`. :wink:

core/src/mender-client.c
```diff
@@ -1050,6 +1055,56 @@ mender_prepare_new_provides(mender_artifact_ctx_t *mender_artifact_ctx, char **n
 
     return MENDER_OK;
 }
+
+static mender_err_t
+mender_check_device_compatibility(mender_artifact_ctx_t *mender_artifact_ctx) {
+
+    /* We need to load the stored provides */
+    mender_key_value_list_t *stored_provides = NULL;
+    if (MENDER_FAIL == mender_storage_get_provides(&stored_provides)) {
```
lluiscampos, line 1064:
```suggestion
    if (MENDER_OK != mender_storage_get_provides(&stored_provides)) {
```

lluiscampos (approved):

core/src/mender-client.c
danielskinstad: `mender_storage_get_provides` can return `MENDER_NOT_FOUND`. With the current design we still need to check if there are unmet depends if there are no provides in the storage.

lluiscampos: Ok, it makes sense.

Alternatively we could have an early check for the Artifact having dependencies at all. If there are none we can skip the check (return ok) and if there are then here we must have provides in the store (or return error). This is just thinking out loud, I think the current code is good as-is

PR #31, opened 2024-08-26
fix: Interpret also `ENOENT` error as not found

lluiscampos:
According to the API docs, `nvs_read` returns the number of bytes read. There is no explicit mention that a return of 0 can be interpret as "not found".

Maybe it worked for some specific platforms if the driver(s) were implemented with this in mind, but generally speaking we should accept also `ENOENT`.


larsewi (approved):
🚀

PR #35, opened 2024-08-31
fix: Workaround I/O errors on some NVS drivers

lluiscampos:
At least for the ESP driver, a `nvs_read` from a NULL pointer gives I/O error, even when `len` is set to 0. My understanding of the API docs is that it should not be the case, but my painfully slow discovery tells otherwise.

Workaround this by allocating 1 byte first for the peek read.

Changelog: None
Ticket: None

### DIFF
--- a/platform/storage/zephyr/nvs/src/mender-storage.c
+++ b/platform/storage/zephyr/nvs/src/mender-storage.c
@@ -50,8 +50,9 @@ static mender_err_t
 nvs_read_alloc(struct nvs_fs *nvs, uint16_t id, void **data, size_t *length) {
     ssize_t ret;
 
-    /* Retrieve length of the data */
-    ret = nvs_read(nvs, id, NULL, 0);
+    /* Peek read to retrieve length of the data */
+    uint8_t byte;
+    ret = nvs_read(nvs, id, &byte, 0);
     if (ret <= 0) {
         return (0 == ret || -ENOENT == ret) ? MENDER_NOT_FOUND : MENDER_FAIL;
     }


platform/storage/zephyr/nvs/src/mender-storage.c
```diff
@@ -50,15 +50,24 @@ static mender_err_t
...
-    ret = nvs_read(nvs, id, NULL, 0);
+    /* Allocate 1 byte fro the first peek read */
+    *data = malloc(1);
+    if (NULL == *data) {
+        mender_log_error("Unable to allocate memory for: %d", id);
+        return MENDER_FAIL;
+    }
+
+    /* Peek read to retrieve length of the data */
+    ret = nvs_read(nvs, id, *data, 0);
```
vpodzime, line 61: I think it's better to do `unsigned char byte` and then `nvs_read(nvs, id, &byte, 0)`; You don't need heap-allocated space to read into, a stack-allocated variable is good enough and way cheaper.

vpodzime (approved):
Looks good to me otherwise. Thanks for the painful hours spent on this for all of us! :beers:

platform/storage/zephyr/nvs/src/mender-storage.c
```diff
@@ -50,15 +50,24 @@ static mender_err_t
 nvs_read_alloc(struct nvs_fs *nvs, uint16_t id, void **data, size_t *length) {
     ssize_t ret;
 
-    /* Retrieve length of the data */
-    ret = nvs_read(nvs, id, NULL, 0);
+    /* Allocate 1 byte fro the first peek read */
```
pasinskim, line 53: `s/fro/from`

platform/storage/zephyr/nvs/src/mender-storage.c
```diff
@@ -50,15 +50,24 @@ static mender_err_t
...
+    *data = malloc(1);
```
pasinskim, line 54: This is pretty interesting. Do you know why we need to do it? I did not see any references to it while reviewing the NVS part.

vpodzime: I believe it's a buggy implementation of the API on Lluis' target board/environment.

lluiscampos: The driver for my specific flash throws "invalid parameter" error when `data` is NULL, even when you ask to read 0 bytes. So yes, I consider it a bug in the driver, really.

platform/storage/zephyr/nvs/src/mender-storage.c
lluiscampos: Of course, thank you.

PR #36, opened 2024-09-02
MEN-7479: Rework artifact_name load and store

danielskinstad:


core/src/mender-api.c
```diff
@@ -272,7 +278,22 @@ api_check_for_deployment_v2(int *status, void *response) {
...
+    if (MENDER_FAIL == mender_storage_get_provides(&provides)) {
+        mender_log_error("Unable to get provides");
+        goto END;
+    }
+    for (mender_key_value_list_t *item = provides; NULL != item; item = item->next) {
+        if (NULL == cJSON_AddStringToObject(json_provides, item->key, item->value)) {
+            mender_log_error("Unable to allocate memory");
+            goto END;
+        }
+    }
```
vpodzime, line 299: We should destroy the list somewhere, right?

core/src/mender-api.c
```diff
@@ -103,11 +103,16 @@ mender_err_t
 mender_api_init(mender_api_config_t *config) {
 
     assert(NULL != config);
-    assert(NULL != config->artifact_name);
     assert(NULL != config->device_type);
     assert(NULL != config->host);
     mender_err_t ret;
 
+    /* Load and set artifact_name here */
+    if (MENDER_OK != mender_storage_get_artifact_name(&artifact_name) && NULL != artifact_name && NULL != artifact_name->value) {
```
vpodzime, line 111: Why do we need `artifact_name` to be a key-value list? We only seem to take the first value anyway and we don't care about key(s).

core/src/mender-client.c
```diff
@@ -805,6 +801,77 @@ mender_client_initialization_work_function(void) {
...
+
+    /* Replace the stored provides with the new provides */
+    if (MENDER_OK != mender_storage_set_provides(new_provides)) {
+        mender_log_error("Unable to set provides");
+        return MENDER_FAIL;
+    }
+#endif /* CONFIG_MENDER_PROVIDES_DEPENDS */
+#endif /* CONFIG_MENDER_FULL_PARSE_ARTIFACT */
+
+    return MENDER_OK;
```
vpodzime, line 872: What does destroy `provides` and `new_provides`?

platform/storage/zephyr/nvs/src/mender-storage.c
```diff
@@ -401,6 +415,52 @@ mender_storage_delete_provides(void) {
...
+
+    assert(NULL != artifact_name);
+
+    char *artifact_name_str = NULL;
+    if (MENDER_OK != mender_utils_key_value_list_to_string(artifact_name, &artifact_name_str)) {
+        return MENDER_FAIL;
+    }
+
+    /* Write artifact_name */
+    if (nvs_write(&mender_storage_nvs_handle, MENDER_STORAGE_NVS_PROVIDES, artifact_name_str, strlen(artifact_name_str) + 1) < 0) {
```
vpodzime, line 429: Please use `checked_nvs_write()`.

vpodzime (commented):
A couple comments/questions. Thanks for shortening the code in the state machine `switch()`!

core/src/mender-api.c
danielskinstad: Yes, you're right. Overlooked it

core/src/mender-client.c
danielskinstad: Nothing, overlooked it

core/src/mender-api.c
danielskinstad: Yeah, it got a bit messy since I tried to follow the way it was done in the c++ client, where we have artifact_name as a part of the provides and they delete it etc., I will rework this - and I think that only the value should be stored :+1:

core/src/mender-api.c
```diff
@@ -86,6 +87,11 @@ static mender_err_t mender_api_websocket_callback(mender_websocket_client_event_
 
 #endif /* CONFIG_MENDER_CLIENT_ADD_ON_TROUBLESHOOT */
 
+/**
+ * @brief Artifact name variable
+ */
+static mender_key_value_list_t *artifact_name = NULL;
+
```
vpodzime, line 94: This belongs to the second commit.

platform/storage/posix/src/mender-storage.c
```diff
@@ -37,11 +37,20 @@
...
 #define MENDER_STORAGE_NVS_PROVIDES        CONFIG_MENDER_STORAGE_PATH "provides.txt"
+#define MENDER_STORAGE_NVS_ARTICACT_NAME   CONFIG_MENDER_STORAGE_PATH "artifact_name.txt"
 
 mender_err_t
 mender_storage_init(void) {
 
-    /* Nothing to do */
+    /* Get artifact name - set to unknown if not found */
+    char *artifact_name = NULL;
+    if (MENDER_NOT_FOUND == mender_storage_get_artifact_name(&artifact_name)) {
```
larsewi, line 47: Instead of writing `"unknown"` to the flash if `MENDER_NOT_FOUND`, can't we just make this function return it? Hence, we would spare the flash for one write?

core/src/mender-api.c
```diff
@@ -315,7 +337,6 @@ api_check_for_deployment_v1(int *status, void *response) {
     char        *path = NULL;
 
     /* Compute path */
-    /* TODO: Retrieve artifact name from store (see ticket MEN-7479) */
```
vpodzime, line 318: This change also belongs into the second commit.

core/src/mender-api.c
```diff
@@ -103,11 +103,15 @@ mender_err_t
 mender_api_init(mender_api_config_t *config) {
 
     assert(NULL != config);
-    assert(NULL != config->artifact_name);
     assert(NULL != config->device_type);
     assert(NULL != config->host);
     mender_err_t ret;
 
+    /* Load and set artifact_name here */
+    if (MENDER_OK != mender_storage_get_artifact_name(&artifact_name) && NULL != artifact_name) {
```
vpodzime, line 111: We typically use parentheses in complex boolean expressions like this.

platform/storage/posix/src/mender-storage.c
```diff
@@ -358,6 +367,31 @@ mender_storage_delete_provides(void) {
 #endif /*CONFIG_MENDER_FULL_PARSE_ARTIFACT*/
 #endif /*CONFIG_MENDER_PROVIDES_DEPENDS*/
 
+mender_err_t
+mender_storage_set_artifact_name(char *artifact_name) {
```
larsewi, line 371:
```suggestion
mender_storage_set_artifact_name(const char *artifact_name) {
```

core/src/mender-client.c
```diff
@@ -344,10 +342,9 @@ mender_client_init(mender_client_config_t *config, mender_client_callbacks_t *ca
         goto END;
     }
     mender_api_config_t mender_api_config = {
-        .artifact_name = mender_client_config.artifact_name,
-        .device_type   = mender_client_config.device_type,
-        .host          = mender_client_config.host,
-        .tenant_token  = mender_client_config.tenant_token,
+        .device_type  = mender_client_config.device_type,
+        .host         = mender_client_config.host,
+        .tenant_token = mender_client_config.tenant_token,
```
vpodzime, line 347: These are not very fortunate formatting rules. Or rather, the enforced application of them...

platform/storage/zephyr/nvs/src/mender-storage.c
```diff
@@ -401,6 +411,41 @@ mender_storage_delete_provides(void) {
 #endif /* CONFIG_MENDER_PROVIDES_DEPENDS */
 #endif /* CONFIG_MENDER_FULL_PARSE_ARTIFACT */
 
+mender_err_t
+mender_storage_set_artifact_name(char *artifact_name) {
```
larsewi, line 415:
```suggestion
mender_storage_set_artifact_name(const char *artifact_name) {
```

platform/storage/zephyr/nvs/src/mender-storage.c
```diff
@@ -401,6 +411,41 @@ mender_storage_delete_provides(void) {
...
+}
+
+mender_err_t
+mender_storage_get_artifact_name(char **artifact_name) {
+
+    assert(NULL != artifact_name);
+    size_t artifact_name_length = 0;
+
+    /* Read artifact_name */
+    mender_err_t ret = nvs_read_alloc(&mender_storage_nvs_handle, MENDER_STORAGE_NVS_PROVIDES, (void **)artifact_name_str, &artifact_name_length);
```
larsewi, line 436: Can we pass `NULL` when we don't care about the length?

core/src/mender-client.c
```diff
@@ -805,6 +801,51 @@ mender_client_initialization_work_function(void) {
     return ret;
 }
 
+static mender_err_t
+mender_commit_artifact_data(void) {
+
+    /* If provides-depends is defined, then `artifact_name` comes from the provides, else
+     * it comes from the deployment data */
```
vpodzime, line 808: The code below doesn't seem to follow this comment, I think it belongs elsewhere.

core/src/mender-client.c
```diff
@@ -1289,10 +1343,12 @@ mender_client_update_work_function(void) {
                         if (MENDER_OK == (ret = mender_check_artifact_requirements(mender_artifact_ctx, deployment))) {
 #ifdef CONFIG_MENDER_PROVIDES_DEPENDS
                             /* Add the new provides to the deployment data (we need the artifact context) */
-                            char *new_provides = NULL;
-                            if (MENDER_OK == (ret = mender_prepare_new_provides(mender_artifact_ctx, &new_provides))) {
+                            char *new_provides  = NULL;
+                            char *artifact_name = NULL;
+                            if (MENDER_OK == (ret = mender_prepare_new_provides(mender_artifact_ctx, &new_provides, &artifact_name))) {
                                 cJSON_AddStringToObject(mender_client_deployment_data, "provides", new_provides);
-                                free(new_provides);
```
vpodzime, line 1295: So what does `free(new_provides)` if we remove it from here?

platform/storage/zephyr/nvs/src/mender-storage.c
```diff
@@ -401,6 +411,41 @@ mender_storage_delete_provides(void) {
...
 
+mender_err_t
+mender_storage_set_artifact_name(char *artifact_name) {
+
+    assert(NULL != artifact_name);
+
+    /* Write artifact_name */
+    if (!checked_nvs_write(&mender_storage_nvs_handle, MENDER_STORAGE_NVS_PROVIDES, artifact_name, strlen(artifact_name) + 1)) {
+        mender_log_error("Unable to write artifact_name");
+        free(artifact_name_str);
```
vpodzime, line 422: No such variable here.

platform/storage/zephyr/nvs/src/mender-storage.c
vpodzime, line 436: No `artifact_name_str` here.

platform/storage/zephyr/nvs/src/mender-storage.c
```diff
@@ -401,6 +411,41 @@ mender_storage_delete_provides(void) {
...
+    if (MENDER_OK != ret) {
+        if (MENDER_NOT_FOUND == ret) {
+            mender_log_info("Provides not available");
+        } else {
+            mender_log_error("Unable to read artifact_name");
+        }
+        return ret;
+    }
+
+    return MENDER_OK;
```
larsewi, line 446: Can consider `switch`
```suggestion
    switch (ret) {
        case MENDER_OK:
            /* all good */
            break;
        case MENDER_NOT_FOUND:
            mender_log_info("Provides not available");
            break;
        default:
            mender_log_error("Unable to read artifact_name");
            break;
    }
    return ret;
```

vpodzime (commented):
A couple things. Thanks for keeping the pace at shortening the big `switch()`! :clap:

core/src/mender-api.c
```diff
@@ -97,11 +103,15 @@ mender_err_t
...
```
larsewi, line 111: We need to `free(artifact_name);` at exit?

platform/storage/zephyr/nvs/src/mender-storage.c
danielskinstad: Not with the current implementation

platform/storage/posix/src/mender-storage.c
danielskinstad: I guess I could make `mender_storage_get_artifact_name` just set artifact_name to unknown on `MENDER_NOT_FOUND`

larsewi: What I mean is that we should avoid unnecessary writes to the flash, so that we don't ware it out

platform/storage/zephyr/nvs/src/mender-storage.c
danielskinstad: Oops, left over from provides

platform/storage/zephyr/nvs/src/mender-storage.c
vpodzime: I'd prefer listing all possible `ret` values in the second `case` and not using `default`.

platform/storage/posix/src/mender-storage.c
```diff
@@ -358,6 +359,39 @@ mender_storage_delete_provides(void) {
...
+        if (MENDER_NOT_FOUND == ret) {
+            *artifact_name = strdup("unknown");
+            return MENDER_OK;
+        } else {
+            mender_log_error("Unable to read artifact_name");
+        }
+        return ret;
+    }
+
+    return MENDER_OK;
```
larsewi, line 392:
```suggestion
        }
        mender_log_error("Unable to read artifact_name");
    }

    return ret;
```

platform/storage/zephyr/nvs/src/mender-storage.c
```diff
@@ -401,6 +402,43 @@ mender_storage_delete_provides(void) {
...
+            *artifact_name = strdup("unknown");
+            return MENDER_OK;
+
+        } else {
+            mender_log_error("Unable to read artifact_name");
+        }
+        return ret;
+    }
+
+    return MENDER_OK;
```
larsewi, line 439:
```suggestion
        }
        mender_log_error("Unable to read artifact_name");
    }

    return ret;
```

platform/storage/zephyr/nvs/src/mender-storage.c
larsewi: This one is not resolved, as far as I can see?

danielskinstad: You're right, guess I didn't commit it, thanks

core/src/mender-api.c
```diff
@@ -337,10 +346,7 @@ api_check_for_deployment_v1(int *status, void *response) {
     char        *path = NULL;
 
     /* Compute path */
-    /* TODO: Retrieve artifact name from store (see ticket MEN-7479) */
-    if (-1
-        == asprintf(
-            &path, MENDER_API_PATH_GET_NEXT_DEPLOYMENT "?artifact_name=%s&device_type=%s", mender_api_config.artifact_name, mender_api_config.device_type)) {
+    if (-1 == asprintf(&path, MENDER_API_PATH_GET_NEXT_DEPLOYMENT "?artifact_name=%s&device_type=%s", artifact_name, mender_api_config.device_type)) {
```
vpodzime, line 349: I think we should add `assert(NULL != artifact_name)` to the beginning of all functions that assume so.

core/src/mender-api.c
```diff
@@ -863,6 +869,7 @@ mender_api_exit(void) {
         free(mender_api_jwt);
         mender_api_jwt = NULL;
     }
+    free(artifact_name);
```
vpodzime, line 872: This seems wrong. The value is loaded in `mender_api_init()`, it should not be destroyed in `mender_api_publish_inventory_data()`.

core/src/mender-client.c
```diff
@@ -805,6 +801,48 @@ mender_client_initialization_work_function(void) {
...
+        return MENDER_FAIL;
+    }
+    mender_utils_free_linked_list(new_provides);
+#endif /* CONFIG_MENDER_PROVIDES_DEPENDS */
+#endif /* CONFIG_MENDER_FULL_PARSE_ARTIFACT */
+
+    if (MENDER_OK != mender_storage_set_artifact_name(json_artifact_name->valuestring)) {
+        mender_log_error("Unable to set artifact name");
+        return MENDER_FAIL;
+    }
```
vpodzime, line 841: I'd do this first to keep the code dealing with `artifact_name` together. But not a biggie.

core/src/mender-client.c
```diff
@@ -1028,7 +1066,7 @@ mender_filter_provides(mender_artifact_ctx_t *mender_artifact_ctx, mender_key_va
 }
 
 static mender_err_t
-mender_prepare_new_provides(mender_artifact_ctx_t *mender_artifact_ctx, char **new_provides) {
+mender_prepare_new_provides(mender_artifact_ctx_t *mender_artifact_ctx, char **new_provides, char **artifact_name) {
 
     assert(NULL != mender_artifact_ctx);
```
vpodzime, line 1071: `assert(NULL != artifact_name)`

core/src/mender-client.c
```diff
@@ -1049,6 +1086,20 @@ mender_prepare_new_provides(mender_artifact_ctx_t *mender_artifact_ctx, char **n
...
+    for (mender_key_value_list_t *item = mender_artifact_ctx->artifact_info.provides; NULL != item; item = item->next) {
+        if (StringEqual("artifact_name", item->key)) {
+            *artifact_name = item->value;
+            break;
+        }
+    }
+
+    if (NULL == *artifact_name) {
+        mender_log_error("No artifact name found in provides");
+        mender_utils_free_linked_list(stored_provides);
```
vpodzime, line 1099: `stored_provides` seems to be leaked by this function unless this `if` is executed.

core/src/mender-client.c
```diff
@@ -1028,7 +1066,7 @@ mender_filter_provides(mender_artifact_ctx_t *mender_artifact_ctx, mender_key_va
 }
 
 static mender_err_t
-mender_prepare_new_provides(mender_artifact_ctx_t *mender_artifact_ctx, char **new_provides) {
+mender_prepare_new_provides(mender_artifact_ctx_t *mender_artifact_ctx, char **new_provides, char **artifact_name) {
```
vpodzime, line 1069: `const char **artifact_name` to signal that it's not going to contain a newly-allocated string.

vpodzime (commented):
A couple things still...

core/src/mender-api.c
danielskinstad: I don't quite get what you mean, it's destroyed in `mender_api_exit()`

core/src/mender-client.c
danielskinstad: `stored_provides` is freed after being combined with `new_provides` in `mender_filter_provides()`

core/src/mender-api.c
vpodzime: oh! GitHub was trying to fool me! It was showing the `mender_api_publish_inventory_data()` function name in the diff header of this chunk! (I double-checked). You're right, sorry for the noise! It might be worth doing `artifact_name = NULL` here, just to make sure some other code doesn't do use-after-free or double-free.

core/src/mender-client.c
vpodzime: Worth a comment.

core/src/mender-client.c
danielskinstad: Yeah, I agree :+1:

core/src/mender-api.c
```diff
@@ -98,11 +103,15 @@ mender_err_t
 mender_api_init(mender_api_config_t *config) {
 
     assert(NULL != config);
-    assert(NULL != config->artifact_name);
     assert(NULL != config->device_type);
     assert(NULL != config->host);
     mender_err_t ret;
 
+    /* Load and set artifact_name here */
+    if ((MENDER_OK != mender_storage_get_artifact_name(&artifact_name)) && (NULL != artifact_name)) {
```
lluiscampos, line 111: This is incorrect. If `provides` contains artifact_name, that one should take precedence. That is why in the acceptance criteria of the task I suggested to have a function that reads the two store keys and merge the results, so that the artifact_name is only one and is consistent after load.

Does it make sense?

core/src/mender-api.c
```diff
@@ -289,8 +299,8 @@ api_check_for_deployment_v2(int *status, void *response) {
     }
 #endif /* CONFIG_MENDER_FULL_PARSE_ARTIFACT */
 #endif /* CONFIG_MENDER_PROVIDES_DEPENDS */
-    /* TODO: Retrieve artifact name from store (see ticket MEN-7479) */
-    if (NULL == cJSON_AddStringToObject(json_provides, "artifact_name", mender_api_config.artifact_name)) {
+
+    if (NULL == cJSON_AddStringToObject(json_provides, "artifact_name", artifact_name)) {
```
lluiscampos, line 303: Relates to my previous comment. The `artifact_name` is expected to be part of the provides already and should not be added here.

core/src/mender-client.c
```diff
@@ -805,6 +801,48 @@ mender_client_initialization_work_function(void) {
...
+
+    if (MENDER_OK != mender_storage_set_artifact_name(json_artifact_name->valuestring)) {
+        mender_log_error("Unable to set artifact name");
+        return MENDER_FAIL;
+    }
+
+#ifdef CONFIG_MENDER_FULL_PARSE_ARTIFACT
+#ifdef CONFIG_MENDER_PROVIDES_DEPENDS
+    /* Get provides from the deployment data */
+    cJSON *provides = cJSON_GetObjectItemCaseSensitive(mender_client_deployment_data, "provides");
```
lluiscampos, line 821: Will the provides from the deployment data contain already `artifact_name`? (I think yes, is it handled in `mender_prepare_new_provides`?) Otherwise we should add it on the fly before storing.

core/src/mender-client.c
```diff
@@ -1305,9 +1358,12 @@ mender_client_update_work_function(void) {
...
 #ifdef CONFIG_MENDER_PROVIDES_DEPENDS
                             /* Add the new provides to the deployment data (we need the artifact context) */
-                            char *new_provides = NULL;
-                            if (MENDER_OK == (ret = mender_prepare_new_provides(mender_artifact_ctx, &new_provides))) {
+                            char       *new_provides  = NULL;
+                            const char *artifact_name = NULL;
+                            if (MENDER_OK == (ret = mender_prepare_new_provides(mender_artifact_ctx, &new_provides, &artifact_name))) {
                                 cJSON_AddStringToObject(mender_client_deployment_data, "provides", new_provides);
+                                /* Replace artifact_name with the one from provides */
+                                cJSON_ReplaceItemInObject(mender_client_deployment_data, "artifact_name", cJSON_CreateString(artifact_name));
```
lluiscampos, line 1366: Why? This seems wrong to me. The whole point of having two is that an Artifact can in theory have a name and also _provide_ an artifact_name.

(I guess I relates to my previous comment - we miss-understood a bit each other here)

lluiscampos (changes requested):
We need to rework this a bit. If you prefer so, we can have a f2f chat for it. Thanks for your patience!

core/src/mender-api.c
danielskinstad: But the artifact_name from provides does take precedence now? This function loads it from the store, and the artifact name stored is either from the deployment, or the one from provides if provides-depends is defined. That is why I overwrite the `artifact_name` key in the deployment data; it is overwritten by the artifact_name from provides. Perhaps it should be a separate key like `provides_artifact_name` ? I didn't see a reason to merge artifact_name from provides into new_provides only to delete it again before storing, if that makes sense.

danielskinstad: Wait, I may have written that comment too fast. It looks like the logic changed a bit after I addressed some comments and made some changes, I'll need to look at it again to see how it works :sweat_smile:

lluiscampos: Discussed f2f and found an agreement. Waiting for the rework :smiley:

lluiscampos (approved):

platform/storage/posix/src/mender-storage.c
```diff
@@ -315,6 +316,38 @@ mender_storage_delete_provides(void) {
...
+}
+
+mender_err_t
+mender_storage_get_artifact_name(char **artifact_name) {
+
+    assert(NULL != artifact_name);
+
+    size_t       artifact_name_length;
+    mender_err_t ret = mender_storage_read_file(MENDER_STORAGE_NVS_ARTICACT_NAME, (void **)artifact_name, &artifact_name_length);
```
vpodzime, line 338: Looks like this will not read nor add the terminating `\0` byte.

platform/storage/zephyr/nvs/src/mender-storage.c
```diff
@@ -286,6 +287,41 @@ mender_storage_delete_provides(void) {
...
+}
+
+mender_err_t
+mender_storage_get_artifact_name(char **artifact_name) {
+
+    assert(NULL != artifact_name);
+    size_t artifact_name_length = 0;
```
vpodzime, line 308: No need to initialize this. (save cycles! :grin: )

vpodzime (approved):
Just one potential issue and a nitpick. No need for a re-review after double-checking or fixing. :)

platform/storage/posix/src/mender-storage.c
danielskinstad: The terminating `\0` byte is added in `mender_storage_read_file()`

PR #38, opened 2024-09-02
Move deployment checking to download state

danielskinstad:


core/src/mender-client.c
```diff
@@ -1203,6 +1203,40 @@ mender_check_artifact_requirements(mender_artifact_ctx_t *mender_artifact_ctx, m
...
+    /* Create deployment data */
+    if (NULL != mender_client_deployment_data) {
+        mender_log_warning("Unexpected stale deployment data");
+        cJSON_Delete(mender_client_deployment_data);
+    }
+    if (NULL == (mender_client_deployment_data = cJSON_CreateObject())) {
+        mender_log_error("Unable to allocate memory");
+        return MENDER_FAIL;
+    }
+    cJSON_AddStringToObject(mender_client_deployment_data, "id", deployment->id);
```
larsewi, line 1233: Unrelated to this PR, but we should probably check the return value of these function calls as they return `NULL` on error. Feel free to fix it if you want :)

larsewi (approved):

core/src/mender-client.c
```diff
@@ -1291,20 +1290,14 @@ mender_client_update_work_function(void) {
...
-    }
     while (MENDER_UPDATE_STATE_END != update_state) {
         switch (update_state) {
             case MENDER_UPDATE_STATE_DOWNLOAD:
                 mender_log_info("Downloading deployment artifact with id '%s', artifact name '%s' and uri '%s'",
                                 deployment->id,
                                 deployment->artifact_name,
                                 deployment->uri);
-                mender_client_publish_deployment_status(deployment_id, MENDER_DEPLOYMENT_STATUS_DOWNLOADING);
+                mender_client_publish_deployment_status(deployment->id, MENDER_DEPLOYMENT_STATUS_DOWNLOADING);
```
vpodzime, line 1309: This was done on purpose to avoid issues when `deployment` is not initialized above. Keep in mind that this `switch()` can jump into an arbitrary branch.

core/src/mender-client.c
```diff
@@ -1203,6 +1203,40 @@ mender_check_artifact_requirements(mender_artifact_ctx_t *mender_artifact_ctx, m
...
+    }
+
+    /* Check if deployment is available */
+    if ((NULL == deployment->id) || (NULL == deployment->artifact_name) || (NULL == deployment->uri) || (NULL == deployment->device_types_compatible)) {
+        mender_log_info("No deployment available");
+        return MENDER_DONE;
+    }
+
+    /* Reset flags */
+    mender_client_deployment_needs_set_pending_image = false;
```
vpodzime, line 1222: I don't think this belongs into this function.

vpodzime (commented):
My gut feeling is that the `deployment` variable should only be used in the `DOWNLOAD` state. Everywhere else we should rely on `mender_client_deployment_data`. (see the comment in the first commit for why)

core/src/mender-client.c
danielskinstad: Ah, yes - I didn't think about that

vpodzime (approved):
Looks good to me.

PR #41, opened 2024-09-02
chore: Remove handling of new_provides from mender_client_authentication_work_function()

vpodzime:
It just doesn't belong in there, it's now handled in the update module state machine as part of the DOWNLOAD and COMMIT states.

Ticket: MEN-7498
Changelog: none

### DIFF
--- a/core/src/mender-client.c
+++ b/core/src/mender-client.c
@@ -842,13 +842,6 @@ mender_client_authentication_work_function(void) {
         }
     }
 
-#ifdef CONFIG_MENDER_FULL_PARSE_ARTIFACT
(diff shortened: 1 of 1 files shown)


vpodzime:
@danielskinstad please double-check this is being properly done in https://github.com/mendersoftware/mender-mcu/blob/feature-update_modules/core/src/mender-client.c#L1322-L1331 and https://github.com/mendersoftware/mender-mcu/blob/feature-update_modules/core/src/mender-client.c#L1405-L1423 instead.

danielskinstad (approved):

PR #43, opened 2024-09-03
Add cJSON Zephyr Module

lluiscampos:


west.yml
```diff
@@ -0,0 +1,13 @@
...
+
+  remotes:
+    - name: mender
+      url-base: https://github.com/mendersoftware
+
+  projects:
+    - name: cjson-zephyr
+      submodules: true
+      remote: mender
+      path: modules/lib/cjson-zephyr
```
vpodzime, line 11: Is this where it goes when fetching/checking out locally?

vpodzime (approved):
Looks good to me.

zephyr/Kconfig
```diff
@@ -424,4 +424,21 @@ if MENDER_MCU_CLIENT
 
     endif
 
+    menu "Optional functionality (ADVANCED)"
+
+        config MENDER_FULL_PARSE_ARTIFACT
+            bool "Full Mender Artifact parse"
+            default y
+            help
+                Do full parsing of the Artifacts allowoing for robustness feature of matching the device type
```
larsewi, line 433:
```suggestion
                Do full parsing of the Artifacts allowing for robustness feature of matching the device type
```

larsewi (approved):

west.yml
lluiscampos: Yes!

It is a bit arbitrary, but in `modules/lib` I found similar libraries like `zcbor`

lluiscampos: But actually, I made a mistake in the final submit: I wanted it to be `modules/lib/cjson` only, without the `-zephyr` suffix which in this context is superfluous. I'll amend this and merge :+1:

zephyr/Kconfig
lluiscampos: :pray:

PR #44, opened 2024-09-03
fix: replaced sscanf with stroul

larsewi:
Replaced multiple calls to `sscanf` with `stroul` in mender artifact parsing, because the calls to `sscanf` resulted in unexplained segmentation faults on some hardware.

### DIFF
--- a/core/src/mender-artifact.c
+++ b/core/src/mender-artifact.c
@@ -17,6 +17,8 @@
  * limitations under the License.
  */
 
+#include <errno.h>
+
 #include "mender-artifact.h"
 #include "mender-log.h"
 
@@ -392,7 +394,17 @@ mender_artifact_parse_tar_header(mender_artifact_ctx_t *ctx) {
     ctx->file.name = tmp;
 
     /* Retrieve file size */
-    sscanf(tar_header->size, "%o", (unsigned int *)&(ctx->file.size));
+    assert(sizeof(size_t) >= sizeof(unsigned long long));
+    char *end_ptr;
+    errno = 0; /* to distinguish between success/failure */
+
+    ctx->file.size = strtoull(tar_header->size, &end_ptr, 8);
+    if ((end_ptr == tar_header->size) /* no conversion */
+        || (0 != errno)) {            /* out of range (for unsigned long) */
+        mender_log_error("Unable to retrieve file size");
+        return MENDER_FAIL;
+    }
+
     ctx->file.index = 0;
 
     /* Shift data in the buffer */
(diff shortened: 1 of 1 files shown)


larsewi:
I'm not sure if I should merge to `main` or `feature-update_modules`?

core/src/mender-artifact.c
```diff
@@ -392,7 +394,15 @@ mender_artifact_parse_tar_header(mender_artifact_ctx_t *ctx) {
     ctx->file.name = tmp;
 
     /* Retrieve file size */
-    sscanf(tar_header->size, "%o", (unsigned int *)&(ctx->file.size));
+    char *end_ptr;
+    errno          = 0; /* to distinguish between success/failure */
+    ctx->file.size = strtoul(tar_header->size, &end_ptr, 8);
```
vpodzime, line 399: Here we are assuming `sizeof(size_t) >= sizeof(unsigned long)`. I guess that's a safe assumption, but an `assert()` would not hurt.

core/src/mender-artifact.c
```diff
@@ -781,18 +791,31 @@ static mender_err_t
...
-    size_t index = 0;
 
-    /* Retrieve payload index */
-    if (1 != sscanf(ctx->file.name, "header.tar/headers/%u/meta-data", (unsigned int *)&index)) {
+    /* Retrieve payload index. We expect "header.tar/headers/%u/meta-data" where
+     * %u is the index. Yes sscanf(3) would be nice, but we've experienced
+     * unexplained segmentation faults on some hardware when using it. */
+    const char *const prefix     = "header.tar/headers/";
+    const size_t      prefix_len = strlen(prefix);
+    if (strncmp(ctx->file.name, prefix, prefix_len) != 0) {
```
vpodzime, line 800: We have `mender_utils_strbeginwith()`.

core/src/mender-artifact.c
vpodzime: The same applies below.

core/src/mender-artifact.c
```diff
@@ -781,18 +791,31 @@ static mender_err_t
...
+    errno              = 0; /* to distinguish between success/failure */
+    const size_t index = strtoul(start_ptr, &end_ptr, 10);
+    if ((end_ptr == start_ptr)              /* no conversion */
+        || (0 != errno)                     /* out of range (for unsigned long) */
+        || (index >= ctx->payloads.size)) { /* index out of bounds */
         mender_log_error("Invalid artifact format");
         return MENDER_FAIL;
     }
 
+    assert(StringEqual(end_ptr, "/meta-data")); /* just one last sanity check */
```
vpodzime, line 817: :+1:

vpodzime (commented):
Looks good to me otherwise.

core/src/mender-artifact.c
lluiscampos, line 399: In theory, the size is up to 11 digits (see tar format [here](https://www.gnu.org/software/tar/manual/html_node/Standard.html), the array is 12 but the last byte will be `\0`). With `unsigned long` is enough, right? Or should we use `long long` to be safe?

(Honest question)

lluiscampos:
> I'm not sure if I should merge to `main` or `feature-update_modules`?

`main`: this fix is not introducing conflicts and we need it for the hardware reference build.

lluiscampos (approved):
Thank you very much

core/src/mender-artifact.c
larsewi: A long can be 4 or 8 Bytes. In case of a 4 Byte unsigned long the largest value it can hold would be
```python3
>>> pow(2, 32) - 1
4294967295
```

larsewi: The largest number 11 digits in base 8 can represent is
```python3
>>> sum = 0
>>> for i in range(10, -1, -1):
...     sum += (7 * 8**i)
... 
>>> sum
8589934591
```

larsewi: So I guess we should can use `strtoull(3)` instead. Nice catch 🚀

larsewi:
>> I'm not sure if I should merge to main or feature-update_modules?

> main: this fix is not introducing conflicts and we need it for the hardware reference build.

Cherry picked to `main` in https://github.com/mendersoftware/mender-mcu/pull/45. Closing this one.

PR #45, opened 2024-09-04
fix: replaced sscanf with strtoull

larsewi:
Replaced multiple calls to `sscanf` with `strtoull` in mender artifact parsing, because the calls to `sscanf` resulted in unexplained segmentation faults on some hardware.

### DIFF
--- a/core/src/mender-artifact.c
+++ b/core/src/mender-artifact.c
@@ -17,6 +17,8 @@
  * limitations under the License.
  */
 
+#include <errno.h>
+
 #include "mender-artifact.h"
 #include "mender-log.h"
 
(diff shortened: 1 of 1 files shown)


core/src/mender-artifact.c
```diff
@@ -392,7 +394,17 @@ mender_artifact_parse_tar_header(mender_artifact_ctx_t *ctx) {
...
 
     /* Retrieve file size */
-    sscanf(tar_header->size, "%o", (unsigned int *)&(ctx->file.size));
+    assert(sizeof(size_t) >= sizeof(unsigned long long));
+    char *end_ptr;
+    errno = 0; /* to distinguish between success/failure */
+
+    ctx->file.size = strtoull(tar_header->size, &end_ptr, 8);
+    if ((end_ptr == tar_header->size) /* no conversion */
+        || (0 != errno)) {            /* out of range (for unsigned long) */
```
vpodzime, line 403: `unsigned long long`

core/src/mender-artifact.c
```diff
@@ -781,18 +793,31 @@ static mender_err_t
...
         return MENDER_FAIL;
     }
-    if (index >= ctx->payloads.size) {
+
+    assert(sizeof(size_t) >= sizeof(unsigned long));
+    const char *start_ptr = ctx->file.name + strlen(prefix);
+    char       *end_ptr;
+    errno = 0; /* to distinguish between success/failure */
+
+    const size_t index = strtoul(start_ptr, &end_ptr, 10);
```
vpodzime, line 811: Not sure if there is any point in preferring `strtoul()` over `strtoull()` here when we have that `assert(sizeof(size_t) >= sizeof(unsigned long long))` and we thus require a big enough `size_t`.

vpodzime (approved):
Looks good to me either way.

lluiscampos (approved):

PR #46, opened 2024-09-04
fix: get_user_provided_keys callback is now optional

larsewi:
The `get_user_provided_keys` callback is now optional. Furthermore, if the callback is set but returns an empty key, the client no longer tries to fix this by generating or loading keys from the store. Now, keys are only generated or loaded from the store when the callback is explicitly set to `NULL`.

### DIFF
--- a/.github/workflows/check_equivalence_tests.sh
+++ b/.github/workflows/check_equivalence_tests.sh
@@ -30,6 +30,7 @@ do
     # Any number of spaces before and after "==" / "!=" is catched
     lines=$(grep "[])a-z][ ]*[!=]=[ ]*[A-Z0-9]" ${source_file} | wc -l)
     if [[ ! $lines -eq 0 ]]; then
+        grep -H -n "[])a-z][ ]*[!=]=[ ]*[A-Z0-9]" ${source_file}
         result="${result}\n${source_file}"
     fi
 done


platform/tls/generic/mbedtls/src/mender-tls.c
```diff
@@ -127,77 +127,90 @@ mender_tls_init(void) {
...
+        &mender_tls_private_key, &mender_tls_private_key_length, &mender_tls_public_key, &mender_tls_public_key_length)) {
+        case MENDER_OK:
+            /* Keys found! */
+            return MENDER_OK;
+        case MENDER_NOT_FOUND:
+            mender_log_debug("Authentication keys not found in store");
+            break;
+        default:
+            mender_log_error("Unable to get authentication keys from store");
+            return MENDER_FAIL;
```
vpodzime, line 193: Please avoid using `default`. By covering all the cases explicitly, the compiler will warn us if there is a new one not covered. It's easy to list them all above the failure block.

.github/workflows/check_equivalence_tests.sh
```diff
@@ -30,6 +30,7 @@ do
     # Any number of spaces before and after "==" / "!=" is catched
     lines=$(grep "[])a-z][ ]*[!=]=[ ]*[A-Z0-9]" ${source_file} | wc -l)
     if [[ ! $lines -eq 0 ]]; then
+        grep -H -n "[])a-z][ ]*[!=]=[ ]*[A-Z0-9]" ${source_file}
```
vpodzime, line 33: :partying_face: :heart: :beers:

vpodzime (approved):
Looks good to me otherwise.

platform/tls/generic/mbedtls/src/mender-tls.c
larsewi: I'm not sure if this is what you meant?

.github/workflows/check_equivalence_tests.sh
larsewi: 🍻

platform/tls/generic/mbedtls/src/mender-tls.c
```diff
@@ -127,77 +127,95 @@ mender_tls_init(void) {
...
         }
 
-        /* Record keys */
         if (MENDER_OK
-            != (ret = mender_storage_set_authentication_keys(
-                    mender_tls_private_key, mender_tls_private_key_length, mender_tls_public_key, mender_tls_public_key_length))) {
-            mender_log_error("Unable to record authentication keys");
-            goto END;
+            != mender_tls_get_authentication_keys(&mender_tls_private_key,
```
lluiscampos, line 166: I think we should split `mender_tls_get_authentication_keys` in two different functions. The API here is very obscure: if you pass user keys it basically converts the format and extracts public key, while if these are null then it generates keys.

We should instead refactor the internals of it, kill this function, and instead us `mender_tls_user_provided_authentication_keys` and `mender_tls_generate_authentication_keys`

lluiscampos (approved):
See my comment below, it is a pre-existing issue so it won't block this PR.

Otherwise :rocket: from my side.

platform/tls/generic/mbedtls/src/mender-tls.c
vpodzime: Yes. Even just this would be good enough for me:
```
case MENDER_DONE:
case MENDER_NOT_IMPLEMENTED:
case MENDER_FAIL:
  mender_log_error(Unable to ...);
  return MENDER_FAIL;
```

vpodzime (approved):
Looks good to me.

platform/tls/generic/mbedtls/src/mender-tls.c
larsewi: Okey, I'll try to do this in a separate PR ;)

PR #47, opened 2024-09-05
fix: fixed error check when reading state

larsewi:
On success, `fread()` return the number of items read. This number equals the number of bytes transferred only when size is 1. Hence, we should check that 1 item is read instead of N Bytes.


vpodzime (approved):
ACK

PR #48, opened 2024-09-05
MEN-7500: Optional default update module

vpodzime:
Plus a couple fixes/improvements.


larsewi (approved):

core/src/mender-client.c
```diff
@@ -363,26 +338,28 @@ mender_client_init(mender_client_config_t *config, mender_client_callbacks_t *ca
...
-    rootfs_image_umod->artifact_type                           = "rootfs-image";
-    rootfs_image_umod->requires_reboot                         = true;
-    rootfs_image_umod->supports_rollback                       = false; /* TODO: support rollback */
+    zephyr_image_umod->callbacks[MENDER_UPDATE_STATE_DOWNLOAD] = &mender_zephyr_image_download_artifact_flash_callback;
+    zephyr_image_umod->callbacks[MENDER_UPDATE_STATE_INSTALL]  = &mender_zephyr_image_set_pending_image;
+    zephyr_image_umod->callbacks[MENDER_UPDATE_STATE_REBOOT]   = &mender_zephyr_image_reboot_callback;
+    zephyr_image_umod->callbacks[MENDER_UPDATE_STATE_FAILURE]  = &mender_zephyr_image_abort_deployment;
+    zephyr_image_umod->artifact_type                           = "zephyr-image";
+    zephyr_image_umod->requires_reboot                         = true;
+    zephyr_image_umod->supports_rollback                       = false; /* TODO: support rollback */
```
lluiscampos, line 354: If all this initialization is happening in `mender-client.c`, it is not clear to me how will I register my custom update module. We should have an entry point in the API for it.

zephyr/Kconfig
```diff
@@ -96,6 +96,16 @@ if MENDER_MCU_CLIENT
...
     endmenu
 
+    menu "Update modules configuration"
+
+        config MENDER_CLIENT_ZEPHYR_IMAGE_UPDATE_MODULE
+            bool "The default Mender zephyr-image update module"
+            default y
+            help
+                The default zephyr-image update module handles full Zephyr images on MCUboot-based devices.
+
```
lluiscampos, line 106: There are a set of dependencies that are only needed if this configuration is on, so we should move here at least:
```
    select BOOTLOADER_MCUBOOT
    select IMG_ENABLE_IMAGE_CHECK
    select IMG_ERASE_PROGRESSIVELY
    select IMG_MANAGER
    select MPU_ALLOW_FLASH_WRITE
    select STREAM_FLASH
```

lluiscampos (changes requested):
Two high level things to discuss :monocle_face:

core/src/mender-client.c
vpodzime: It's there -- [`mender_client_register_update_module()`](https://github.com/mendersoftware/mender-mcu/blob/feature-update_modules/include/mender-client.h#L81). You have to initialize the update module yourself, of course.

zephyr/Kconfig
vpodzime: Are you sure the `MPU_ALLOW_FLASH_WRITE` and `STREAM_FLASH` are not required for NVS?

lluiscampos: > Are you sure the `MPU_ALLOW_FLASH_WRITE` 

This one I copied my mistake.

> and `STREAM_FLASH` are not required for NVS?

It is [not required](https://docs.zephyrproject.org/latest/kconfig.html#CONFIG_STREAM_FLASH), as far as I can tell.

core/src/mender-client.c
lluiscampos: Yet we are still assigning all the properties here, like the artifact_type, the callbacks, etc.

Let me comment this from a different point of view: I would expect `mender-client.c` to know nothing about `zephyr-image` updates, so no `mender_update_module_t *zephyr_image_umod` declared here nor any code under `CONFIG_MENDER_ZEPHYR_IMAGE_UPDATE_MODULE`. Isn't that possible? And does it make sense to you?

vpodzime: oh, right, so you mean it shouldn't be in `mender-client.c` but rather in the code that initializes the client. That makes sense, but having it here means a user code that wants to use the `zephyr-image` type doesn't have to worry about update modules at all. I don't know if that is desirable or not. I can definitely at least move this code into a function in `mender-zephyr-image-update-module.c/h` and then it can be called from one place or another.

lluiscampos: > oh, right, so you mean it shouldn't be in `mender-client.c` but rather in the code that initializes the client. 

Yes! Sorry

> That makes sense, but having it here means a user code that wants to use the `zephyr-image` type doesn't have to worry about update modules at all. I don't know if that is desirable or not.

The thing is, that I would treat "zephyr-image" Update Module any different than the rest. We develop it ourselves and can have an easier KConfig/CMake integration than the rest, but code wise it should be clear that it is just one update module and showcase everything that you need to do in order to create your own.

> I can definitely at least move this code into a function in `mender-zephyr-image-update-module.c/h` and then it can be called from one place or another.

That sounds good to me :+1:

lluiscampos (approved):
Thank you very much for following up on the feedback. I think it looks good!

PR #49, opened 2024-09-06
Moved pending deployment

danielskinstad:
Question: do we need to check anything else, and is it in the correct place? Is comparing the artifact type from the deployment data necessary?

### DIFF
--- a/core/src/mender-client.c
+++ b/core/src/mender-client.c
@@ -792,82 +792,6 @@ mender_client_authentication_work_function(void) {
         }
     }
 
-    /* Check if deployment is pending */
-    if (NULL != mender_client_deployment_data) {
-
-        /* Retrieve deployment data */
-        cJSON *json_id = NULL;
-        if (NULL == (json_id = cJSON_GetObjectItemCaseSensitive(mender_client_deployment_data, "id"))) {
-            mender_log_error("Unable to get ID from the deployment data");
-            goto RELEASE;
-        }
-        char *id;
-        if (NULL == (id = cJSON_GetStringValue(json_id))) {
-            mender_log_error("Unable to get ID from the deployment data");
-            goto RELEASE;
-        }
-        cJSON *json_artifact_name = NULL;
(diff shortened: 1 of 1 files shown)


core/src/mender-client.c
```diff
@@ -1092,6 +1092,11 @@ mender_client_check_deployment(mender_api_deployment_data_t **deployment_data) {
     cJSON_AddStringToObject(mender_client_deployment_data, "artifact_name", deployment->artifact_name);
     cJSON_AddArrayToObject(mender_client_deployment_data, "types");
 
+    assert(NULL != mender_client_deployment_data);
```
vpodzime, line 1095: This is guaranteed by the `if` a couple lines above, right?

vpodzime (approved):
Looks good to me otherwise.

vpodzime:
> Is comparing the artifact type from the deployment data necessary?

No, I didn't quite understand what the purpose of this was. We definitely need to make sure the artifact type is supported by some available update module after a reboot, but that should/does happen in a different place.


> do we need to check anything else, and is it in the correct place?

As suspicious as your change looks I believe it's correct. The logic should jump into a the update function which should either continue or abort an update/deployment or check for a new one.

core/src/mender-client.c
danielskinstad: Yep

vpodzime:
> @vpodzime Where is going to happen? Do we have a task covering that? Otherwise I'd say we need to add it in this PR (the JIRA description was not very clear to me)

We should do it [here](https://github.com/mendersoftware/mender-mcu/blob/feature-update_modules/core/src/mender-client.c#L1197), I think. But let me create a ticket for it, we need to also properly handle the case when there is a new deployment with an artifact of an unsupported type. The current code would jump into the `FAILURE` state and try to call a callback of a `NULL` module.

vpodzime:
> But let me create a ticket for it

https://northerntech.atlassian.net/browse/MEN-7524

PR #50, opened 2024-09-09
Allow authentication failure if there is a pending deployment

danielskinstad:
Ticket: MEN-7490
Changelog: Title

### DIFF
--- a/core/src/mender-client.c
+++ b/core/src/mender-client.c
@@ -579,11 +579,23 @@ mender_client_work_function(void) {
     if (MENDER_OK != (ret = mender_client_network_connect())) {
         goto END;
     }
+
+    /* Flag to allow deployment to continue in spite of a failed authentication */
+    bool allow_authentication_failure = false;
+
     /* Intentional pass-through */
     if (MENDER_CLIENT_STATE_AUTHENTICATION == mender_client_state) {
         /* Perform authentication with the server */
         if (MENDER_DONE != (ret = mender_client_authentication_work_function())) {
-            goto RELEASE;
+            /* Check if there is a deployment in progress */
+            if (NULL != mender_client_deployment_data) {
+                /* Authentication failed with pending deployment */
+                mender_log_info("Authentication failed, continuing with deployment");
+                allow_authentication_failure = true;
+            } else {
+                /* Authentication failed and no pending deployment */
+                goto RELEASE;
+            }
         }
         /* Update work period */
         if (MENDER_OK != (ret = mender_scheduler_work_set_period(mender_client_work_handle, mender_client_config.update_poll_interval))) {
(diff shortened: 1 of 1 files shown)


core/src/mender-client.c
```diff
@@ -633,22 +633,36 @@ mender_client_work_function(void) {
...
     /* Intentional pass-through */
     if (MENDER_CLIENT_STATE_AUTHENTICATION == mender_client_state) {
         /* Perform authentication with the server */
         if (MENDER_DONE != (ret = mender_client_authentication_work_function())) {
-            goto RELEASE;
+            /* Check if there is a deployment in progress */
+            if (NULL != mender_client_deployment_data) {
+                /* Authentication failed with pending deployment */
+                mender_log_info("Authentication failed, continuing with deployment");
+                allow_authentication_failure = true;
```
vpodzime, line 648: @lluiscampos, is this OK for now? IIRC, you talked to Kristian about resuming.

vpodzime (approved):
Looks good to me, but we may need a bigger discussion here. Thanks!

core/src/mender-client.c
vpodzime: There's also the other alternative -- an in-progress deployment that fixes network/authentication (thinking about new/expired certificates, etc.).

kacf: I'm not sure I understand exactly what "pending deployment" means in this context, but the Golang and C++ clients do an authentication after everything has been installed, and booted, but before committing. In Update Module/state script terms, it is right before `ArtifactCommit_Enter`. If this fails it is counted as an error, and the update is rolled back.

Keep in mind that it has to be a complete authentication from scratch, it can not simply reuse the JWT token. Otherwise the device risks getting locked out after 7 days, which happened to some of our customers in the past.

vpodzime: @kacf my thinking was focused on something like the following scenario:
1. the device downloads an update
2. the device switches to the `INSTALL` state
3. the device loses power
4. the device fails to authenticate after the power loss
Should we continue with the update with the downloaded artifact? Or should we start over? Or should we even reboot?

vpodzime: > [This graph](https://docs.mender.io/3.7/artifact-creation/create-a-custom-update-module#power-loss) might be helpful to visualize what we're talking about

Ahhhh, another graph! :grin: But thanks for sharing the link, it is definitely useful. And it's part of the documentation so unless we want add an exception there, we should try to follow the same flow in the MCU client.

Here the biggest aspect of the changes is that without any, the code just skips entering the update module state machine completely. So instead of doing a proper rollback using an (or rather **the**) update module, it just remains stuck in the authenticating state. Thus we need to be able to skip the authentication here, or even network initialization, call the `mender_client_update_work_function()` and let the update module state machine deal with the current state.

PR #52, opened 2024-09-09
feat: added deployment data version number and store counter

larsewi:
- Created deployment data management API.
- Added deployment data version number which allows us to
  add/remove/modify fields in the future.
- Added the deployment data store counter field that prevents infinite
  reboots.
- Deployment data is now validated when reading or writing from store.
- Also did some preparation for adding the update state field later in
  MEN-7515.

### DIFF
--- a/core/src/mender-client.c
+++ b/core/src/mender-client.c
@@ -27,6 +27,7 @@
 #include "mender-tls.h"
 #include "mender-update-module.h"
 #include "mender-utils.h"
+#include "mender-deployment-data.h"
 
 #ifdef CONFIG_MENDER_CLIENT_INVENTORY
 #include "mender-inventory.h"
@@ -116,9 +117,10 @@ static const char *update_state_str[N_MENDER_UPDATE_STATES + 1] = {
 static bool mender_client_network_connected = false;
 
 /**
- * @brief Deployment data (ID, artifact name and payload types), used to report deployment status after rebooting
+ * @brief Deployment data. Used to track progress of an update, so that the
+ *        operation can resume or roll back across reboots
  */
-static cJSON *mender_client_deployment_data = NULL;
+static mender_deployment_data_t *mender_client_deployment_data = NULL;
 
 /**
  * @brief Mender client update modules list
@@ -445,11 +447,8 @@ mender_client_exit(void) {
     mender_client_config.tenant_token                 = NULL;
     mender_client_config.authentication_poll_interval = 0;
     mender_client_config.update_poll_interval         = 0;
-
-    if (NULL != mender_client_deployment_data) {
-        cJSON_Delete(mender_client_deployment_data);
-        mender_client_deployment_data = NULL;
-    }
+    mender_delete_deployment_data(mender_client_deployment_data);
+    mender_client_deployment_data = NULL;
 
     if (NULL != mender_update_modules_list) {
         for (size_t update_module_index = 0; update_module_index < mender_update_modules_count; update_module_index++) {
@@ -491,8 +490,7 @@ mender_client_work_function(void) {
 static mender_err_t
 mender_client_initialization_work_function(void) {
 
-    char        *storage_deployment_data = NULL;
-    mender_err_t ret                     = MENDER_DONE;
+    mender_err_t ret = MENDER_DONE;
 
     /* Retrieve or generate authentication keys */
     if (MENDER_OK != (ret = mender_tls_init_authentication_keys(mender_client_callbacks.get_user_provided_keys, mender_client_config.recommissioning))) {
@@ -501,21 +499,11 @@ mender_client_initialization_work_function(void) {
     }
 
     /* Retrieve deployment data if it is found (following an update) */
-    if (MENDER_OK != (ret = mender_storage_get_deployment_data(&storage_deployment_data))) {
+    if (MENDER_OK != (ret = mender_get_deployment_data(&mender_client_deployment_data))) {
         if (MENDER_NOT_FOUND != ret) {
             mender_log_error("Unable to get deployment data");
             goto REBOOT;
         }
-    }
-
-    if (NULL != storage_deployment_data) {
-        if (NULL == (mender_client_deployment_data = cJSON_Parse(storage_deployment_data))) {
-            mender_log_error("Unable to parse deployment data");
-            free(storage_deployment_data);
-            ret = MENDER_FAIL;
-            goto REBOOT;
-        }
-        free(storage_deployment_data);
     }
 
     mender_log_info("Initialization done");
@@ -545,13 +533,15 @@ REBOOT:
 static mender_err_t
 mender_commit_artifact_data(void) {
 
-    cJSON *json_artifact_name = cJSON_GetObjectItemCaseSensitive(mender_client_deployment_data, "artifact_name");
-    if (NULL == json_artifact_name) {
+    assert(NULL != mender_client_deployment_data);
+
+    const char *artifact_name;
+    if (MENDER_OK != mender_deployment_data_get_artifact_name(mender_client_deployment_data, &artifact_name)) {
         mender_log_error("Unable to get artifact name from the deployment data");
         return MENDER_FAIL;
     }
 
-    if (MENDER_OK != mender_storage_set_artifact_name(json_artifact_name->valuestring)) {
+    if (MENDER_OK != mender_storage_set_artifact_name(artifact_name)) {
         mender_log_error("Unable to set artifact name");
         return MENDER_FAIL;
     }
@@ -559,15 +549,15 @@ mender_commit_artifact_data(void) {
 #ifdef CONFIG_MENDER_FULL_PARSE_ARTIFACT
 #ifdef CONFIG_MENDER_PROVIDES_DEPENDS
     /* Get provides from the deployment data */
-    cJSON *provides = cJSON_GetObjectItemCaseSensitive(mender_client_deployment_data, "provides");
-    if (NULL == provides) {
+    const char *provides;
+    if (MENDER_OK != mender_deployment_data_get_provides(mender_client_deployment_data, &provides)) {
         mender_log_error("Unable to get new_provides from the deployment data");
         return MENDER_FAIL;
     }
 
     /* Parse provides */
     mender_key_value_list_t *new_provides = NULL;
-    if (MENDER_OK != mender_utils_string_to_key_value_list(provides->valuestring, &new_provides)) {
+    if (MENDER_OK != mender_utils_string_to_key_value_list(provides, &new_provides)) {
         mender_log_error("Unable to parse provides from the deployment data");
         return MENDER_FAIL;
     }
@@ -878,19 +868,14 @@ mender_client_check_deployment(mender_api_deployment_data_t **deployment_data) {
     /* Create deployment data */
     if (NULL != mender_client_deployment_data) {
         mender_log_warning("Unexpected stale deployment data");
-        cJSON_Delete(mender_client_deployment_data);
+        mender_delete_deployment_data(mender_client_deployment_data);
     }
-    if (NULL == (mender_client_deployment_data = cJSON_CreateObject())) {
-        mender_log_error("Unable to allocate memory");
+    if (MENDER_OK
+        != (mender_create_deployment_data(
+            deployment->id, deployment->artifact_name, NULL /* We will populate later */, NULL /* TODO: MEN-7515 */, &mender_client_deployment_data))) {
+        /* Error already logged */
         return MENDER_FAIL;
     }
-    cJSON_AddStringToObject(mender_client_deployment_data, "id", deployment->id);
-    cJSON_AddStringToObject(mender_client_deployment_data, "artifact_name", deployment->artifact_name);
-    cJSON_AddArrayToObject(mender_client_deployment_data, "types");
-
-    assert(NULL != cJSON_GetObjectItem(mender_client_deployment_data, "id"));
-    assert(NULL != cJSON_GetObjectItem(mender_client_deployment_data, "artifact_name"));
-    assert(NULL != cJSON_GetObjectItem(mender_client_deployment_data, "types"));
 
     return MENDER_OK;
 }
@@ -903,19 +888,15 @@ mender_client_update_work_function(void) {
     mender_artifact_ctx_t *mender_artifact_ctx = NULL;
 
     /* Check for deployment */
-    mender_api_deployment_data_t *deployment              = NULL;
-    char                         *storage_deployment_data = NULL;
-    mender_update_state_t         update_state            = MENDER_UPDATE_STATE_DOWNLOAD;
-    const char                   *deployment_id           = NULL;
+    mender_api_deployment_data_t *deployment    = NULL;
+    mender_update_state_t         update_state  = MENDER_UPDATE_STATE_DOWNLOAD;
+    const char                   *deployment_id = NULL;
 
     /* reset the currently used update module */
     mender_update_module = NULL;
 
-    {
-        cJSON *deployment_id_j = cJSON_GetObjectItem(mender_client_deployment_data, "id");
-        if (NULL != deployment_id_j) {
-            deployment_id = deployment_id_j->valuestring;
-        }
+    if (NULL != mender_client_deployment_data) {
+        mender_deployment_data_get_id(mender_client_deployment_data, &deployment_id);
     }
 
     {
@@ -1001,9 +982,13 @@ mender_client_update_work_function(void) {
                             char       *new_provides  = NULL;
                             const char *artifact_name = NULL;
                             if (MENDER_OK == (ret = mender_prepare_new_provides(mender_artifact_ctx, &new_provides, &artifact_name))) {
-                                cJSON_AddStringToObject(mender_client_deployment_data, "provides", new_provides);
+                                if (MENDER_OK != (ret = mender_deployment_data_set_provides(mender_client_deployment_data, new_provides))) {
+                                    mender_log_error("Failed to set deployment data provides");
+                                }
                                 /* Replace artifact_name with the one from provides */
-                                cJSON_ReplaceItemInObject(mender_client_deployment_data, "artifact_name", cJSON_CreateString(artifact_name));
+                                else if (MENDER_OK != (ret = mender_deployment_data_set_artifact_name(mender_client_deployment_data, artifact_name))) {
+                                    mender_log_error("Failed to set deployment data artifact name");
+                                }
                                 free(new_provides);
                             } else {
                                 mender_log_error("Unable to prepare new provides");
@@ -1047,10 +1032,7 @@ mender_client_update_work_function(void) {
                 mender_client_publish_deployment_status(deployment_id, MENDER_DEPLOYMENT_STATUS_REBOOTING);
 
                 /* Save deployment data to publish deployment status after rebooting */
-                if (NULL == (storage_deployment_data = cJSON_PrintUnformatted(mender_client_deployment_data))) {
-                    mender_log_error("Unable to serialize deployment data for saving");
-                    ret = MENDER_FAIL;
-                } else if (MENDER_OK != (ret = mender_storage_set_deployment_data(storage_deployment_data))) {
+                if (MENDER_OK != (ret = mender_set_deployment_data(mender_client_deployment_data))) {
                     mender_log_error("Unable to save deployment data");
                 }
                 if ((MENDER_OK == ret) && (NULL != mender_update_module->callbacks[update_state])) {
@@ -1168,13 +1150,8 @@ mender_client_update_work_function(void) {
 END:
     /* Release memory */
     deployment_destroy(deployment);
-    if (NULL != storage_deployment_data) {
-        free(storage_deployment_data);
-    }
-    if (NULL != mender_client_deployment_data) {
-        cJSON_Delete(mender_client_deployment_data);
-        mender_client_deployment_data = NULL;
-    }
+    mender_delete_deployment_data(mender_client_deployment_data);
+    mender_client_deployment_data = NULL;
     mender_artifact_release_ctx(mender_artifact_ctx);
 
     return ret;
@@ -1184,7 +1161,6 @@ static mender_err_t
 mender_client_download_artifact_callback(char *type, cJSON *meta_data, char *filename, size_t size, void *data, size_t index, size_t length) {
 
     assert(NULL != type);
-    cJSON       *json_types;
     mender_err_t ret = MENDER_FAIL;
 
 #if CONFIG_MENDER_LOG_LEVEL >= MENDER_LOG_LEVEL_INF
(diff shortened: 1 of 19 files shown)


core/src/mender-client.c
```diff
@@ -1237,10 +1220,7 @@ mender_client_update_work_function(void) {
                 mender_client_publish_deployment_status(deployment_id, MENDER_DEPLOYMENT_STATUS_REBOOTING);
 
                 /* Save deployment data to publish deployment status after rebooting */
-                if (NULL == (storage_deployment_data = cJSON_PrintUnformatted(mender_client_deployment_data))) {
-                    mender_log_error("Unable to serialize deployment data for saving");
-                    ret = MENDER_FAIL;
-                } else if (MENDER_OK != (ret = mender_storage_set_deployment_data(storage_deployment_data))) {
+                if (MENDER_OK != mender_storage_set_deployment_data(mender_client_deployment_data)) {
```
danielskinstad, line 1223:
```suggestion
                if (MENDER_OK != (ret  = mender_storage_set_deployment_data(mender_client_deployment_data))) {
```
I think we need to assign the return value to `ret` here.

pasinskim:
Overall it looks good but I have one comment. Initially the interface (or SDK, public API) was JSON agnostic. We've been passing strings in places like this one for example `mender_err_t mender_storage_set_deployment_data(cJSON *deployment_data);`. Now you changed it to `cJSON` which means binding to the cJSON library. Is that what we wanted to do? What if user wants to use another JSON library.

vpodzime:
> Now you changed it to `cJSON` which means binding to the cJSON library. Is that what we wanted to do? What if user wants to use another JSON library.

This should be fine once we have the generic JSON interface. However, it will mean our JSON interface would have to be part of the SDK/API. I think the best approach here is to make the type opaque and have a constructor.

core/src/mender-utils.c
```diff
@@ -653,3 +653,55 @@ mender_utils_compare_wildcard(const char *str, const char *wildcard_str, bool *m
...
+
+    static struct key_and_type field[] = {
+        { .key = "version", .type = cJSON_IsNumber },                      /* So we can modify fields later */
+        { .key = "id", .type = cJSON_IsString },                           /* Deployment identifier */
+        { .key = "artifact_name", .type = cJSON_IsString },                /* Name of artifact */
+        { .key = "payload_types", .type = cJSON_IsArray },                 /* Types of payloads embedded in artifact */
+        { .key = "provides", .type = cJSON_IsString },                     /* Artifact provides filtered on clears provides */
+        /* { .key = "name", .type = cJSON_IsString }, */                   /* TODO: Implement (see MEN-7515) */
+        /* { .key = "reboot_requested", .type = cJSON_IsString }, */       /* TODO: Implement */
+        /* { .key = "supports_rollback", .type = cJSON_IsString }, */      /* TODO: Implement */
```
lluiscampos, line 679: Based on Kristian consultation in Slack I think we leave this ones out. We might find a use case where this suddenly is necessary, but then we can just add them.

include/mender-storage.h
```diff
@@ -27,6 +27,11 @@ extern "C" {
 #include "mender-utils.h"
 #include "mender-update-module.h"
 
+/**
+ * @brief Deployment data version number
+ */
+#define MENDER_STORAGE_DEPLOYMENT_DATA_VERSION 1
```
lluiscampos, line 33: I know I [said](https://github.com/mendersoftware/mender-mcu/pull/39#discussion_r1743399477) to put it in this file, but I meant is "as encapsulated as possible", which in this PR means `mender-storage.c`. Makes sense? At the time as we were discussing in the `mender-client.c` I only thought about header files

platform/storage/posix/src/mender-storage.c
```diff
@@ -151,24 +151,101 @@ mender_storage_delete_authentication_keys(void) {
...
-mender_storage_set_deployment_data(char *deployment_data) {
+mender_storage_set_deployment_data(cJSON *deployment_data) {
     assert(NULL != deployment_data);
-    size_t deployment_data_length = strlen(deployment_data);
 
-    if (MENDER_OK != mender_storage_write_file(MENDER_STORAGE_NVS_DEPLOYMENT_DATA, deployment_data, deployment_data_length)) {
+    /* Add/update version number. We always want to store the deployment data in
+     * the format of the latest and greatest version. By updating the version
+     * number here and validating the data below, we make sure to enforce this.
+     */
```
lluiscampos, line 160: Right now `mender_storage_set_deployment_data` is only called when rebooting, while for every other state transition we do `mender_storage_save_update_state`. As we have plans to merge them, let me comment a concern:

If we blindly update the version (when saving the update_state) then we have issues like:
* device uses v1
* reboots into a newer Mender that upgrades the data to v2 in the first state transition
* An error occurs, a rollback is needed
* reboots into the old Mender that does not understand the data

It is not a fatal issue because we would just clean up the deployment and no inform the backend, but the device would be alive.

However, because dealing with migrations is complex, I would for now make the code to only understand one version, no upgrade no nothing. If it has a different version than the one we support, we fail the update at any time. Whenever we have to introduce a v2 then we plan a proper migration step.

lluiscampos (commented):
I have not looked into the code details. See few comments below.

> (...) I think the best approach here is to make the type opaque and have a constructor.

I agree with this.

core/src/mender-client.c
```diff
@@ -585,10 +585,8 @@ mender_client_exit(void) {
     mender_scheduler_mutex_give(mender_client_network_mutex);
     mender_scheduler_mutex_delete(mender_client_network_mutex);
     mender_client_network_mutex = NULL;
-    if (NULL != mender_client_deployment_data) {
-        cJSON_Delete(mender_client_deployment_data);
-        mender_client_deployment_data = NULL;
-    }
+    cJSON_Delete(mender_client_deployment_data);
+    mender_client_deployment_data = NULL;
```
vpodzime, line 589: Time for `FREE_AND_NULL(what, destructor)`?

core/src/mender-utils.c
```diff
@@ -653,3 +653,55 @@ mender_utils_compare_wildcard(const char *str, const char *wildcard_str, bool *m
...
+    /* When we introduce new versions of deployment data in the future, the
+     * checks below should vary based on the version passed as an argument to
+     * this function. */
+
+    struct key_and_type {
+        const char *const key;
+        cJSON_bool (*type)(const cJSON *const);
+    };
+
+    static struct key_and_type field[] = {
```
vpodzime, line 671: `const`?

vpodzime: And `fields`? The `sizeof(field)` below is quite confusing otherwise.

core/src/mender-utils.c
```diff
@@ -653,3 +653,55 @@ mender_utils_compare_wildcard(const char *str, const char *wildcard_str, bool *m
 
     return MENDER_OK;
 }
+
+bool
+mender_utils_validate_deployment_data(const cJSON *deployment_data, int version) {
```
vpodzime, line 658: I'd use an `enum` for `version`.

platform/storage/posix/src/mender-storage.c
```diff
@@ -151,24 +151,101 @@ mender_storage_delete_authentication_keys(void) {
...
+    } else {
+        /* One would expect cJSON_AddItemToObject above to replace the value on
+         * already existing keys. However, it creates another entry with a
+         * duplicate key, which is invalid JSON. So we have to be careful. */
+        if (!cJSON_ReplaceItemInObject(deployment_data, "version", item)) {
+            mender_log_error("Unable to allocate memory");
+            cJSON_Delete(item);
+            return MENDER_FAIL;
+        }
+    }
```
vpodzime, line 181: An alternative would be to always `cJSON_DeleteItemFromObject()` (which does nothing in case it's not there) and then add it.

platform/storage/posix/src/mender-storage.c
```diff
@@ -151,24 +151,101 @@ mender_storage_delete_authentication_keys(void) {
...
+        case MENDER_OK:
+            break;
+        case MENDER_NOT_FOUND:
+            mender_log_debug("Deployment data not available");
+            return ret;
+        case MENDER_DONE:
+        case MENDER_NOT_IMPLEMENTED:
+        case MENDER_FAIL:
+            mender_log_error("Unable to read deployment data");
+            return ret;
```
vpodzime, line 228: Shouldn't we always return `MENDER_FAIL` in this last case? The caller of this function should not bother about the various failure cases, right?

platform/storage/zephyr/nvs/src/mender-storage.c
```diff
@@ -180,34 +180,100 @@ mender_storage_delete_authentication_keys(void) {
...
+        mender_log_error("Invalid deployment data");
+        return MENDER_FAIL;
+    }
+
+    /* Compose JSON string */
+    char *json_str;
+    if (NULL == (json_str = cJSON_PrintUnformatted(deployment_data))) {
+        mender_log_error("Unable to compose deployment data");
+        return MENDER_FAIL;
+    }
```
vpodzime, line 224: Everything above this line in this function is platform-agnostic, right? Can we put the code into a common place, please?

core/src/mender-client.c
```diff
@@ -1072,17 +1072,13 @@ mender_client_check_deployment(mender_api_deployment_data_t **deployment_data) {
         mender_log_warning("Unexpected stale deployment data");
         cJSON_Delete(mender_client_deployment_data);
     }
-    if (NULL == (mender_client_deployment_data = cJSON_CreateObject())) {
+    if ((NULL == (mender_client_deployment_data = cJSON_CreateObject()))
+        || (NULL == cJSON_AddStringToObject(mender_client_deployment_data, "id", deployment->id))
+        || (NULL == cJSON_AddStringToObject(mender_client_deployment_data, "artifact_name", deployment->artifact_name))
+        || (NULL == cJSON_AddArrayToObject(mender_client_deployment_data, "payload_types"))) {
```
vpodzime, line 1078: (pre-existing) Hmm, thinking about how we add/replace the `version` field -- shouldn't we do the same trick here for these fields as well? Or are they guaranteed to not be present in the deployment data?

core/src/mender-client.c
```diff
@@ -1139,7 +1139,7 @@ mender_client_update_work_function(void) {
     {
         cJSON *deployment_id_j = cJSON_GetObjectItem(mender_client_deployment_data, "id");
         if (NULL != deployment_id_j) {
-            deployment_id = deployment_id_j->valuestring;
+            deployment_id = cJSON_GetStringValue(deployment_id_j);
```
vpodzime, line 1142: This is questionable, I think. The internal data structure is not really internal, it's part of the header and thus a part of the public API. A function call comes with an overhead. Accessing the field directly skips any type checking, but it's guaranteed to be a JSON string, right?

vpodzime (commented):
Looks good to me otherwise.

core/src/mender-client.c
```diff
@@ -1383,30 +1356,21 @@ mender_client_download_artifact_callback(char *type, cJSON *meta_data, char *fil
...
+    if (MENDER_OK != mender_deployment_data_get_artifact_name(mender_client_deployment_data, &artifact_name)) {
         mender_log_error("Unable to get artifact name from the deployment data");
         goto END;
     }
 
     /* Invoke update module download callback */
-    struct mender_update_download_state_data_s download_state_data = { id, artifact_name, type, meta_data, filename, size, data, index, length, false };
-    mender_update_state_data_t                 state_data          = { .download_state_data = &download_state_data };
+    struct mender_update_download_state_data_s download_state_data
+        = { (char *)id, (char *)artifact_name, type, meta_data, filename, size, data, index, length, false };
```
larsewi, line 1372: @vpodzime should these be `const char *` in the `mender_update_download_state_data_s` struct?

vpodzime: Yes, we don't want the struct to modify or own those strings. :+1:

core/src/mender-utils.c
larsewi: I changed it to `double` since it is now hidden behind an API and I wanted to avoid type casting. Do you still think this should be `enum`?

platform/storage/posix/src/mender-storage.c
larsewi: This is no longer relevant, as mender-storage.c is effectively untouched now. But yes, I agree 👍

core/src/mender-deployment-data.c
```diff
@@ -0,0 +1,311 @@
...
+
+    /* Check if max state data store count is reached */
+    cJSON *item = cJSON_GetObjectItem(*deployment_data, KEY_STATE_DATA_STORE_COUNT);
+    assert(item != NULL); /* Validation above should have catched this already */
+    if (cJSON_GetNumberValue(item) >= MAX_STATE_DATA_STORE_COUNT) {
+        mender_log_error("Reached max state data store count");
+        cJSON_Delete(*deployment_data);
+        *deployment_data = NULL;
+        return MENDER_FAIL;
+    }
```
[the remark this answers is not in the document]
larsewi: Makes sense :)

include/mender-deployment-data.h
```diff
@@ -0,0 +1,93 @@
...
+ */
+mender_err_t __mender_deployment_data_set_string(mender_deployment_data_t *deployment_data, const char *key, const char *str);
+
+#define mender_deployment_data_set_artifact_name(deployment_data, artifact_name) __mender_deployment_data_set_string(deployment_data, "artifact_name", artifact_name)
+#define mender_deployment_data_set_provides(deployment_data, provides) __mender_deployment_data_set_string(deployment_data, "provides", provides)
+#define mender_deployment_data_set_id(deployment_data, id) __mender_deployment_data_set_string(deployment_data, "id", id)
+#define mender_deployment_data_set_state(deployment_data, state) __mender_deployment_data_set_string(deployment_data, "state", state)
+
+mender_err_t mender_deployment_data_append_payload_type(mender_deployment_data_t *deployment_data, const char *payload_type);
+
```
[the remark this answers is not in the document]
larsewi: 🍻

core/src/mender-client.c
```diff
@@ -1153,12 +1135,8 @@ mender_client_update_work_function(void) {
         continue;                                                                                \
     }
 
-    {
-        cJSON *deployment_id_j = cJSON_GetObjectItem(mender_client_deployment_data, "id");
-        if (NULL != deployment_id_j) {
-            deployment_id = deployment_id_j->valuestring;
-        }
-    }
+    mender_deployment_data_get_id(mender_client_deployment_data, &deployment_id);
```
danielskinstad, line 1138: `mender_client_deployment_data` can be NULL here, which will make the `assert(NULL != deployment_data)` in `__mender_deployment_data_get_string()` fail

core/src/mender-deployment-data.c
```diff
@@ -0,0 +1,311 @@
...
+ *       constant as a double to avoid type casting.
+ */
+#define DEPLOYMENT_DATA_VERSION 1.0
+
+/**
+ * @brief Deployment data version number.
+ * @note cJSON stores numbers as double, so we might as well define this
+ *       constant as a double to avoid type casting.
+ */
+#define MAX_STATE_DATA_STORE_COUNT 10.0 /* TODO: What should this constant be? */
```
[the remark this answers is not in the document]
larsewi: Changed it to 50 for now. But we can experiment later

core/src/mender-client.c
larsewi: Added if guard around it. Not quite sure if it is a proper solution. At least it crashed for @danielskinstad before, and does not crash for him now.

core/src/mender-client.c
larsewi: I would expect `FREE_AND_NULL()` to call `free()`. Should there be two macros? E.g. `FREE_AND_NULL(what)` and `DESTROY_AND_NULL(destructor, what)`

core/src/mender-deployment-data.c
```diff
@@ -0,0 +1,309 @@
...
+
+        /* Make sure the field has correct type */
+        if (!fields[i].type(item)) {
+            mender_log_debug("Bad type for key '%s' in deployment data", fields[i].key);
+            return false;
+        }
+
+        /* Check version compatibility */
+        if (StringEqual(fields[i].key, __MENDER_DEPLOYMENT_DATA_KEY_VERSION)) {
+            /* Double is returned below, so we need to cast it to int */
```
larsewi, line 82: Remove this comment

core/src/mender-client.c
```diff
@@ -670,30 +660,29 @@ mender_client_initialization_work_function(void) {
 
 static mender_err_t
 mender_commit_artifact_data(void) {
-
-    cJSON *json_artifact_name = cJSON_GetObjectItemCaseSensitive(mender_client_deployment_data, "artifact_name");
-    if (NULL == json_artifact_name) {
+    const char *artifact_name;
+    if (MENDER_OK != mender_deployment_data_get_artifact_name(mender_client_deployment_data, &artifact_name)) {
```
vpodzime, line 664: I'd add `assert(m_c_d_d != NULL)` to the beginning of the function. I know it's not strictly necessary, but it might prevent strange issues in testing.

core/src/mender-client.c
```diff
@@ -1111,9 +1092,13 @@ mender_client_update_work_function(void) {
                             char       *new_provides  = NULL;
                             const char *artifact_name = NULL;
                             if (MENDER_OK == (ret = mender_prepare_new_provides(mender_artifact_ctx, &new_provides, &artifact_name))) {
-                                cJSON_AddStringToObject(mender_client_deployment_data, "provides", new_provides);
+                                if (MENDER_OK != (ret = mender_deployment_data_set_provides(mender_client_deployment_data, new_provides))) {
+                                    mender_log_error("Failed to set deployment data provides");
+                                }
                                 /* Replace artifact_name with the one from provides */
-                                cJSON_ReplaceItemInObject(mender_client_deployment_data, "artifact_name", cJSON_CreateString(artifact_name));
+                                if (MENDER_OK != (ret = mender_deployment_data_set_artifact_name(mender_client_deployment_data, artifact_name))) {
```
vpodzime, line 1099: This potentially overrides a previous failure. I think we only want to do this if the above succeeds.

core/src/mender-client.c
vpodzime: Are you changing this?

vpodzime (commented):
Looks good to me otherwise.

include/mender-deployment-data.h
```diff
@@ -0,0 +1,173 @@
...
+#ifndef __MENDER_DEPLOYMENT_DATA_H__
+#define __MENDER_DEPLOYMENT_DATA_H__
+
+#ifdef __cplusplus
+extern "C" {
+#endif /* __cplusplus */
+
+#include "mender-utils.h" /* mender_err_t */
+#include "cJSON.h"        /* cJSON */
```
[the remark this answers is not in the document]
larsewi: Haha 😆

include/mender-deployment-data.h
```diff
@@ -0,0 +1,173 @@
...
+
+#define __MENDER_DEPLOYMENT_DATA_KEY_VERSION                "version"
+#define __MENDER_DEPLOYMENT_DATA_KEY_ID                     "id"
+#define __MENDER_DEPLOYMENT_DATA_KEY_ARTIFACT_NAME          "artifact_name"
+#define __MENDER_DEPLOYMENT_DATA_KEY_PAYLOAD_TYPES          "payload_types"
+#define __MENDER_DEPLOYMENT_DATA_KEY_PROVIDES               "provides"
+#define __MENDER_DEPLOYMENT_DATA_KEY_NAME                   "name"
+#define __MENDER_DEPLOYMENT_DATA_KEY_STATE_DATA_STORE_COUNT "state_data_store_count"
```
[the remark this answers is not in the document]
larsewi: They used to be a part of `mender-deployment-data.c`. But @lluiscampos had a valid point in https://github.com/mendersoftware/mender-mcu/pull/52#discussion_r1756638661 to also use them in the header. I was thinking the underscores would be a good compromise between the two. But I don't have a strong opinion. So I will remove them ;)

include/mender-deployment-data.h
```diff
@@ -0,0 +1,173 @@
...
+
+#include "mender-utils.h" /* mender_err_t */
+#include "cJSON.h"        /* cJSON */
+
+#define __MENDER_DEPLOYMENT_DATA_KEY_VERSION                "version"
+#define __MENDER_DEPLOYMENT_DATA_KEY_ID                     "id"
+#define __MENDER_DEPLOYMENT_DATA_KEY_ARTIFACT_NAME          "artifact_name"
+#define __MENDER_DEPLOYMENT_DATA_KEY_PAYLOAD_TYPES          "payload_types"
+#define __MENDER_DEPLOYMENT_DATA_KEY_PROVIDES               "provides"
+#define __MENDER_DEPLOYMENT_DATA_KEY_NAME                   "name"
```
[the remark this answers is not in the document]
larsewi: Name is what it is called in C++ client.

core/src/mender-client.c
larsewi: Will do ;)

core/src/mender-deployment-data.c
```diff
@@ -0,0 +1,309 @@
...
+        return MENDER_FAIL;
+    }
+
+    /* Check if max state data store count is reached */
+    cJSON *item = cJSON_GetObjectItemCaseSensitive(deployment_data, __MENDER_DEPLOYMENT_DATA_KEY_STATE_DATA_STORE_COUNT);
+    assert(NULL != item); /* Validation above should have catched this already */
+    if (MAX_STATE_DATA_STORE_COUNT <= cJSON_GetNumberValue(item)) {
+        mender_log_error("Reached max state data store count");
+        cJSON_Delete(deployment_data);
+        deployment_data = NULL;
```
larsewi, line 110:
```suggestion
```

core/src/mender-deployment-data.c
```diff
@@ -0,0 +1,311 @@
...
+            mender_log_debug("Bad type for key '%s' in deployment data", fields[i].key);
+            return false;
+        }
+
+        /* Check version compatibility */
+        if (StringEqual(fields[i].key, MENDER_DEPLOYMENT_DATA_KEY_VERSION)) {
+            /* Trying to avoid floating-point precision errors */
+            const double delta = (DEPLOYMENT_DATA_VERSION > cJSON_GetNumberValue(item)) ? DEPLOYMENT_DATA_VERSION - cJSON_GetNumberValue(item)
+                                                                                        : cJSON_GetNumberValue(item) - DEPLOYMENT_DATA_VERSION;
+            if (delta > 0.01) {
```
vpodzime, line 85: :joy: did you also get nervous? :wink:

include/mender-deployment-data.h
```diff
@@ -0,0 +1,174 @@
...
+
+#include "mender-utils.h"
+#include "cJSON.h"
+
+#define MENDER_DEPLOYMENT_DATA_KEY_VERSION                "version"
+#define MENDER_DEPLOYMENT_DATA_KEY_ID                     "id"
+#define MENDER_DEPLOYMENT_DATA_KEY_ARTIFACT_NAME          "artifact_name"
+#define MENDER_DEPLOYMENT_DATA_KEY_PAYLOAD_TYPES          "payload_types"
+#define MENDER_DEPLOYMENT_DATA_KEY_PROVIDES               "provides"
+#define MENDER_DEPLOYMENT_DATA_KEY_NAME                   "name"
```
vpodzime, line 35: I still don't like this name of the field. If it's a state I don't see why we should call it "name".

include/mender-flash.h
```diff
@@ -33,7 +33,7 @@ extern "C" {
  * @param handle Handle of the deployment to be used with mender flash functions
  * @return MENDER_OK if the function succeeds, error code otherwise
  */
-mender_err_t mender_flash_open(char *name, size_t size, void **handle);
+mender_err_t mender_flash_open(const char *name, size_t size, void **handle);
```
vpodzime, line 36: The commit message should mention these as well.

vpodzime (approved):
Looks good to me otherwise. Great work!

core/src/mender-deployment-data.c
larsewi: Yes, you made me nervous 😉

include/mender-deployment-data.h
larsewi: It is what it is called in the C++ client. But I don't mind changing it. Should we call it `"state_name"`, or just `"state"`?

danielskinstad: FYI, I renamed it to `state` in my task that builds on this, but I guess both works

larsewi: Okey, then we do `state` ;)

vpodzime (approved):
Thanks!

lluiscampos (approved):
Well done.

PR #53, opened 2024-09-10
feat: Switch auto-generation of crypto keys to ECDSA

lluiscampos:
We expect users to provide their own keys through the callback API, but for the cases when we auto-generate switch to shorter and faster ECDSA.

### DIFF
--- a/platform/tls/generic/mbedtls/src/mender-tls.c
+++ b/platform/tls/generic/mbedtls/src/mender-tls.c
@@ -26,7 +26,7 @@
 #include <mbedtls/error.h>
 #endif /* MBEDTLS_ERROR_C */
 #include <mbedtls/pk.h>
-#include <mbedtls/rsa.h>
+#include <mbedtls/ecdsa.h>
 #include <mbedtls/x509.h>
 #include "mender-log.h"
 #include "mender-storage.h"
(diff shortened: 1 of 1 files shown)


platform/tls/generic/mbedtls/src/mender-tls.c
```diff
@@ -406,13 +407,25 @@ mender_tls_generate_authentication_keys(mbedtls_pk_context *pk_context) {
...
-    if (0 != (ret = mbedtls_pk_setup(pk_context, mbedtls_pk_info_from_type(MBEDTLS_PK_RSA)))) {
+    if (0 != (ret = mbedtls_pk_setup(pk_context, mbedtls_pk_info_from_type(MBEDTLS_PK_ECKEY)))) {
         LOG_MBEDTLS_ERROR("Unable to setup pk", ret);
         goto END;
     }
 
+    /* Find a supported curve */
+    for (curve_info = mbedtls_ecp_curve_list(); MBEDTLS_ECP_DP_NONE != curve_info->grp_id; curve_info++) {
+        if (1 == mbedtls_ecdsa_can_do(curve_info->grp_id)) {
+            mender_log_debug("Found ECDSA curve: %s", curve_info->name);
```
vpodzime, line 418: *supported* ECDSA curve

vpodzime (approved):
I wouldn't say ECDSA is *safer*. No need to have that word in the commit message. Looks good to me otherwise.

PR #54, opened 2024-09-10
feat: Support for two CA certificates

lluiscampos:
The secondary is optional, and it is meant for the cases where the Artifacts are downloaded from a different server (or rather, a server that needs a different CA)

### DIFF
--- a/zephyr/Kconfig
+++ b/zephyr/Kconfig
@@ -268,11 +268,23 @@ if MENDER_MCU_CLIENT
 
         menu "Network options (ADVANCED)"
 
-            config MENDER_NET_CA_CERTIFICATE_TAG
-                int "CA certificate tag"
+            config MENDER_NET_CA_CERTIFICATE_TAG_PRIMARY
+                int "Primary CA certificate tag for Server"
                 default 1
                 help
-                    A security tag that ROOT CA server credential will be referenced with, see tls_credential_add.
+                    A primary security tag that ROOT CA server credential will be referenced with, typically used to authenicate the Mender Server. See tls_credential_add.
+
+            config MENDER_NET_CA_CERTIFICATE_TAG_SECONDARY
+                int "Secondary CA certificate tag for Artifacts"
+                default 2
+                help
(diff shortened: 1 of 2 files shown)


zephyr/Kconfig
```diff
@@ -268,11 +268,23 @@ if MENDER_MCU_CLIENT
...
+            config MENDER_NET_CA_CERTIFICATE_TAG_PRIMARY
+                int "Primary CA certificate tag for Server"
                 default 1
                 help
-                    A security tag that ROOT CA server credential will be referenced with, see tls_credential_add.
+                    A primary security tag that ROOT CA server credential will be referenced with, typically used to authenicate the Mender Server. See tls_credential_add.
+
+            config MENDER_NET_CA_CERTIFICATE_TAG_SECONDARY
+                int "Secondary CA certificate tag for Artifacts"
+                default 2
```
vpodzime, line 279: Does this mean it can be set to 3 or generally any number? Is that a supported case?

vpodzime (commented):
It's hard to believe this is all we need. But it looks good tome.

zephyr/Kconfig
vpodzime: > The `2` here does not mean "two" certificates, but just the id of the second one

I know. But it's part of the Kconfig, right? Basically what I'm asking about is if we can force this to be `2` or unset. Otherwise the user can set this to `3` or even `-1`, I guess...?

lluiscampos: The user should be able to set it to any _valid_ value, as we cannot enforce this. For example they might have tag 1 for a telemetry server or whatever, and Mender will then use tags 2 and 3.

Does it make sense?

vpodzime: > Does it make sense?

Yup! I thought it would be as you describe but without seeing where those tags are coming from (and what they are), I wasn't sure.

vpodzime (approved):

larsewi (approved):

PR #55, opened 2024-09-10
fix: Fix asserts from Artifact parsing

lluiscampos:
On ESP32 Extensa, the sizeof size_t is 4 bytes, which was making the first assert(s) fail. Use `unsigned long` instead.

And the second assert(s) where also crashing, I think it's  because `strcmp` will not stop comparing until getting the null byte...

### DIFF
--- a/core/src/mender-artifact.c
+++ b/core/src/mender-artifact.c
@@ -394,13 +394,13 @@ mender_artifact_parse_tar_header(mender_artifact_ctx_t *ctx) {
     ctx->file.name = tmp;
 
     /* Retrieve file size */
-    assert(sizeof(size_t) >= sizeof(unsigned long long));
+    assert(sizeof(size_t) >= sizeof(unsigned long));
     char *end_ptr;
     errno = 0; /* to distinguish between success/failure */
 
-    ctx->file.size = strtoull(tar_header->size, &end_ptr, 8);
+    ctx->file.size = strtoul(tar_header->size, &end_ptr, 8);
     if ((end_ptr == tar_header->size) /* no conversion */
-        || (0 != errno)) {            /* out of range (for unsigned long long) */
+        || (0 != errno)) {            /* out of range (for unsigned long) */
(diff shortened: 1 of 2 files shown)


core/src/mender-artifact.c
```diff
@@ -394,13 +394,13 @@ mender_artifact_parse_tar_header(mender_artifact_ctx_t *ctx) {
     ctx->file.name = tmp;
 
     /* Retrieve file size */
-    assert(sizeof(size_t) >= sizeof(unsigned long long));
+    assert(sizeof(size_t) >= sizeof(unsigned long));
     char *end_ptr;
     errno = 0; /* to distinguish between success/failure */
 
     ctx->file.size = strtoull(tar_header->size, &end_ptr, 8);
```
vpodzime, line 401: `strtoull()` still used here.

vpodzime (commented):
Looks good to me otherwise.

core/src/mender-artifact.c
```diff
@@ -803,20 +803,21 @@ mender_artifact_read_meta_data(mender_artifact_ctx_t *ctx) {
...
-        || (0 != errno)                     /* out of range (for unsigned long long) */
+        || (0 != errno)                     /* out of range (for unsigned long) */
         || (index >= ctx->payloads.size)) { /* index out of bounds */
         mender_log_error("Invalid artifact format");
         return MENDER_FAIL;
     }
 
-    assert(StringEqual(end_ptr, "/meta-data")); /* just one last sanity check */
+    assert(NULL != end_ptr);
+    assert(0 == strncmp(end_ptr, "/meta-data", 10)); /* just one last sanity check */
```
vpodzime, line 820: Let's please add `StringEqualN()` and use it here and below. It could be based on a potentially faster `memcmp()` since we apparently cannot rely on `str*` functions stopping on `\0`-bytes anyway...

vpodzime (approved):
Looks good to me either way.

larsewi (approved):

PR #58, opened 2024-09-10
Final touches for stable runtime :see_no_evil: 

lluiscampos:
Both changes are provisional and we have tasks to follow-up. See individual commits :popcorn:

### DIFF
--- a/core/src/mender-client.c
+++ b/core/src/mender-client.c
@@ -80,12 +80,6 @@ typedef enum {
 static mender_client_state_t mender_client_state = MENDER_CLIENT_STATE_INITIALIZATION;
 
 /**
- * @brief Counter and mutex for the management of network connect/release callbacks
- */
-static uint8_t mender_client_network_count = 0;
-static void   *mender_client_network_mutex = NULL;
-
-/**
  * @brief Deployment data (ID, artifact name and payload types), used to report deployment status after rebooting
  */
 static cJSON *mender_client_deployment_data = NULL;
@@ -114,11 +108,6 @@ static void                           *mender_client_artifact_types_mutex = NULL
 static mender_addon_instance_t **mender_client_addons_list  = NULL;
 static size_t                    mender_client_addons_count = 0;
 static void                     *mender_client_addons_mutex = NULL;
-
-/**
- * @brief Mender client work handle
- */
-static void *mender_client_work_handle = NULL;
 
 /**
  * @brief Flash handle used to store temporary reference to write rootfs-image data
@@ -259,6 +248,9 @@ mender_client_init(mender_client_config_t *config, mender_client_callbacks_t *ca
 
     mender_client_config.artifact_name = config->artifact_name;
     mender_client_config.device_type   = config->device_type;
+
+    mender_log_info("Artifact name: [%s]; device type: [%s]", mender_client_config.artifact_name, mender_client_config.device_type);
+
     if ((NULL != config->host) && (strlen(config->host) > 0)) {
         mender_client_config.host = config->host;
     } else {
@@ -298,7 +290,7 @@ mender_client_init(mender_client_config_t *config, mender_client_callbacks_t *ca
     memcpy(&mender_client_callbacks, callbacks, sizeof(mender_client_callbacks_t));
 
     /* Initializations */
-    if (MENDER_OK != (ret = mender_scheduler_init())) {
+    if (MENDER_OK != (ret = mender_scheduler_alt_work_create(mender_client_work_function, mender_client_config.update_poll_interval))) {
         mender_log_error("Unable to initialize scheduler");
         goto END;
     }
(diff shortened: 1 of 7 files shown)


core/src/mender-client.c
```diff
@@ -1165,17 +1009,30 @@ mender_client_update_work_function(void) {
...
+    mender_log_info("Getting device type");
+
     /* Retrieve device type from artifact */
     const char *device_type_artifact = NULL;
     if (MENDER_OK != (ret = mender_artifact_get_device_type(mender_artifact_ctx, &device_type_artifact))) {
         mender_log_error("Unable to get device type from artifact");
-        mender_client_publish_deployment_status(deployment->id, MENDER_DEPLOYMENT_STATUS_FAILURE);
-        if (mender_client_deployment_needs_set_pending_image) {
-            mender_flash_abort_deployment(mender_client_flash_handle);
+        if (true == only_for_testing) {
```
larsewi, line 1020: No need to compare boolean value with `true`

core/src/mender-client.c
```diff
@@ -1165,17 +1009,30 @@ mender_client_update_work_function(void) {
...
+            mender_log_error("Will be using PROJECT_NAME as device type for testing purposes");
+            char *device_type_test = malloc(strlen("PROJECT_NAME") + 1);
+            strcpy(device_type_test, "PROJECT_NAME");
```
larsewi, line 1023:
```suggestion
            char *device_type_test = strdup("PROJECT_NAME");
```

core/src/mender-client.c
```diff
@@ -1272,10 +1132,13 @@ mender_client_update_work_function(void) {
...
         mender_client_deployment_data = NULL;
     }
-    mender_artifact_release_ctx(mender_artifact_ctx);
+    mender_log_info("Releasing artifact context");
+    // we will be restarting anyways
+    // mender_artifact_release_ctx(mender_artifact_ctx);
 
     /* Check if the system must restart following downloading the deployment */
     if (true == mender_client_deployment_needs_restart) {
+        mender_log_info("All righty; get it done baby!");
```
larsewi, line 1141: 😆

larsewi (approved):
🚀

core/src/mender-client.c
```diff
@@ -655,65 +519,51 @@ mender_client_exit(void) {
...
+            }
+            mender_client_state = MENDER_CLIENT_STATE_AUTHENTICATED;
+            break;
+        case MENDER_CLIENT_STATE_AUTHENTICATED:
+            mender_log_info("work function; authenticated");
+            ret = mender_client_update_work_function();
+            break;
+        default:
+            mender_log_info("work function; default");
+            break;
```
vpodzime, line 563: Please avoid using `default`, we should explicitly cover all possible values of the enum.

core/src/mender-client.c
```diff
@@ -655,65 +519,51 @@ mender_client_exit(void) {
...
-    mender_client_network_release();
+    switch (mender_client_state) {
+        case MENDER_CLIENT_STATE_INITIALIZATION:
+            mender_log_info("work function; initialization");
+            if (MENDER_DONE != (ret = mender_client_initialization_work_function())) {
+                return ret;
+            }
+            /* Update client state */
+            mender_client_state = MENDER_CLIENT_STATE_AUTHENTICATION;
+            break;
```
vpodzime, line 548: why `break` here? Can't we just continue without waiting for the next execution of this function hitting this `switch()` again?

core/src/mender-client.c
vpodzime: Plus check that it was actually allocated as is not `NULL`.

core/src/mender-client.c
vpodzime: :joy:

vpodzime (commented):
I wish I could say it looks *good* to me, but it's a great step for the Mender-kind! Hence :+1: :beers:

core/src/mender-client.c
lluiscampos: Actually, I have to remove this. Thanks

core/src/mender-client.c
lluiscampos: Thanks

core/src/mender-client.c
lluiscampos: @pasinskim words...

core/src/mender-client.c
lluiscampos: Yes, I'll do that.

core/src/mender-client.c
lluiscampos: (I mean, you can see that it is just set above to `true` and never modified...`

core/src/mender-client.c
lluiscampos: This code is removed. See above comment.

core/src/mender-client.c
lluiscampos: We are already going through all of them. Do you mean completely removing the `default` or adding an assert to make sure we never have an untreated case?

(this is a question in general, as I know you've been commenting on these before)

lluiscampos: > (...) completely removing the `default` (...)

I did this

core/src/mender-client.c
```diff
@@ -1272,10 +1117,13 @@ mender_client_update_work_function(void) {
         cJSON_Delete(mender_client_deployment_data);
         mender_client_deployment_data = NULL;
     }
-    mender_artifact_release_ctx(mender_artifact_ctx);
+    mender_log_info("Releasing artifact context");
+    // we will be restarting anyways
```
pasinskim, line 1121: Maybe you can consider adding a `TODO` comment that we should check the context releasing (deinitialization; cleanup) before the restart.

pasinskim:
Nice. Thanks for tidying it up!

core/src/mender-client.c
```diff
@@ -376,12 +340,6 @@ mender_client_register_artifact_type(char *type,
     mender_client_artifact_type_t **tmp;
     mender_err_t                    ret;
```
lluiscampos, line 341: Note to self,
```suggestion
    mender_err_t                    ret = MENDER_OK;
```

Here we are missing initialization. Please **review all of these**

core/src/mender-client.c
vpodzime: Yes, completely removing, the compiler will then warn us if there is an unhandled case.

vpodzime (approved):

platform/scheduler/posix/src/mender-scheduler.c
```diff
@@ -458,3 +458,22 @@ mender_scheduler_work_queue_thread(void *arg) {
...
     pthread_exit(NULL);
 }
+
+/**
+ * @brief Dummy implementation
+ */
+mender_err_t
+mender_scheduler_alt_work_create(mender_scheduler_alt_work_function_t func, int32_t interval) {
+    (void)func;
+    (void)interval;
```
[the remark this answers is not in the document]
lluiscampos: [Not really](https://gitlab.com/Northern.tech/Mender/mender-mcu/-/jobs/7813220664) :sweat_smile:

PR #61, opened 2024-09-12
Handle unsupported artifact types

danielskinstad:
Handle unsupported artifact types when we 1) load a saved state and 2) when we get a new deployment.

### DIFF
--- a/core/src/mender-client.c
+++ b/core/src/mender-client.c
@@ -1034,10 +1034,26 @@ mender_client_update_work_function(void) {
     mender_update_module = NULL;
 
     {
+        cJSON *deployment_id_j = cJSON_GetObjectItem(mender_client_deployment_data, "id");
+        if (NULL != deployment_id_j) {
+            deployment_id = deployment_id_j->valuestring;
+        }
+    }
+
+    {
         char *artifact_type;
         if (MENDER_OK == (ret = mender_storage_get_update_state(&update_state, &artifact_type))) {
             mender_log_debug("Resuming from state %s", update_state_str[update_state]);
             mender_update_module = mender_client_get_update_module(artifact_type);
+            if (NULL == mender_update_module) {
+                /* The artifact_type from the saved state does not match any update module */
+                mender_log_error("No update module found for artifact type '%s'", artifact_type);
+                mender_client_publish_deployment_status(deployment_id, MENDER_DEPLOYMENT_STATUS_FAILURE);
+                mender_storage_delete_deployment_data();
+                mender_storage_delete_update_state();
+                free(artifact_type);
+                goto END;
+            }
             free(artifact_type);
         }
     }
@@ -1068,12 +1084,6 @@ mender_client_update_work_function(void) {
         continue;                                                                                \
     }
 
-    {
-        cJSON *deployment_id_j = cJSON_GetObjectItem(mender_client_deployment_data, "id");
-        if (NULL != deployment_id_j) {
-            deployment_id = deployment_id_j->valuestring;
-        }
-    }
     while (MENDER_UPDATE_STATE_END != update_state) {
         switch (update_state) {
             case MENDER_UPDATE_STATE_DOWNLOAD:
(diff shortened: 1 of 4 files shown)


include/mender-storage.h
```diff
@@ -89,6 +89,12 @@ mender_err_t mender_storage_save_update_state(mender_update_state_t state, const
  */
 mender_err_t mender_storage_get_update_state(mender_update_state_t *state, char **artifact_type);
 
+/**
+ * @brief Delete update module state
+ * @see mender_storage_delete_update_state()
```
larsewi, line 94: > @see mender_storage_delete_update_state()

What is the purpose of this?

core/src/mender-client.c
```diff
@@ -1080,11 +1080,27 @@ mender_client_update_work_function(void) {
     /* reset the currently used update module */
     mender_update_module = NULL;
 
+    {
+        cJSON *deployment_id_j = cJSON_GetObjectItem(mender_client_deployment_data, "id");
+        if (NULL != deployment_id_j) {
+            deployment_id = deployment_id_j->valuestring;
```
larsewi, line 1086: `cJSON_GetStringValue()` will also check that the type is actually string. However, it does provides some overhead. And you are just moving code around. So feel free to dismiss ;)

platform/storage/posix/src/mender-storage.c
```diff
@@ -259,6 +259,16 @@ mender_storage_get_update_state(mender_update_state_t *state, char **artifact_ty
     fclose(f);
     return ret;
 }
+mender_err_t
+mender_storage_delete_update_state(void) {
+    /* Delete update state */
```
larsewi, line 264: These comments kind of states the obvious. But it does not hurt.

larsewi (approved):

include/mender-storage.h
danielskinstad: Oh, I didn't even see that, I copied the comment from save_update_state without reading

core/src/mender-client.c
```diff
@@ -1080,11 +1080,27 @@ mender_client_update_work_function(void) {
...
+        }
```
lluiscampos, line 1087: Pre-existing,

What happens if there is no deployment id? Can we continue at all? (We won't be able to send status to the server, for example)

lluiscampos (approved):
It looks good!

vpodzime (commented):
A couple things.

core/src/mender-client.c
```diff
@@ -1080,11 +1080,27 @@ mender_client_update_work_function(void) {
     /* reset the currently used update module */
     mender_update_module = NULL;
 
+    {
+        cJSON *deployment_id_j = cJSON_GetObjectItem(mender_client_deployment_data, "id");
```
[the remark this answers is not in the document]
danielskinstad: I have just moved it around, so there's not really a difference from before - but nothing happens if you pass a nullpointer here, it just returns NULL, maybe I am missing your point

platform/storage/zephyr/nvs/src/mender-storage.c
```diff
@@ -225,6 +225,18 @@ mender_storage_delete_deployment_data(void) {
     return MENDER_OK;
 }
 
+mender_err_t
+mender_storage_delete_update_state(void) {
+
+    /* Delete update state */
+    if (0 != nvs_delete(&mender_storage_nvs_handle, MENDER_STORAGE_NVS_UPDATE_STATE)) {
```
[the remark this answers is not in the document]
danielskinstad: ah, i didn't catch that :+1:

core/src/mender-client.c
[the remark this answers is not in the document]
vpodzime: > nothing happens if you pass a nullpointer here, it just returns NULL, maybe I am missing your point

oh, no problem then. I thought it would crash. :cool:

vpodzime (approved):
Looks good to me.

PR #62, opened 2024-09-13
chore: Use configurable buffer for logging and check logging errors.

pasinskim:
Logger improvements.

### DIFF
--- a/platform/log/zephyr/src/mender-log.c
+++ b/platform/log/zephyr/src/mender-log.c
@@ -33,31 +33,44 @@ mender_log_init(void) {
     return MENDER_OK;
 }
 
+/* IMPORTANT: Please note that the default size of internal Zephyr logger buffer is 1024 bytes; if we need log buffer size 
+              to be more than that there is CONFIG_LOG_BUFFER_SIZE option allowing to extend the buffer size. */
+#define LOG_MESSAGE_MAX_SIZE_BYTES (128)
+
 mender_err_t
 mender_log_print(uint8_t level, const char *filename, const char *function, int line, char *format, ...) {
 
-    (void)function;
-    char log[256] = { 0 };
+    (void)filename;
 
-    /* Format message */
     va_list args;
     va_start(args, format);
-    vsnprintf(log, sizeof(log), format, args);
+
+    char log_buff[LOG_MESSAGE_MAX_SIZE_BYTES];
+    int  ret = vsnprintf(log_buff, LOG_MESSAGE_MAX_SIZE_BYTES, format, args);
+
     va_end(args);
+
+    if (ret < 0) {
+        LOG_ERR("logger error: log message formatting failed: %d", ret);
+        return MENDER_FAIL;
+    } else if (ret >= LOG_MESSAGE_MAX_SIZE_BYTES) {
+        /* Log message is too long; add ... at the end */
+        memset(&log_buff[LOG_MESSAGE_MAX_SIZE_BYTES - 4], '.', 3);
+    }
 
     /* Switch depending log level */
     switch (level) {
         case MENDER_LOG_LEVEL_ERR:
-            LOG_ERR("%s (%d): %s", filename, line, log);
+            LOG_ERR("%s", log_buff);
             break;
         case MENDER_LOG_LEVEL_WRN:
-            LOG_WRN("%s (%d): %s", filename, line, log);
+            LOG_WRN("%s", log_buff);
             break;
         case MENDER_LOG_LEVEL_INF:
-            LOG_INF("%s (%d): %s", filename, line, log);
+            LOG_INF("%s", log_buff);
             break;
         case MENDER_LOG_LEVEL_DBG:
-            LOG_DBG("%s (%d): %s", filename, line, log);
+            LOG_DBG("%s (%d): %s", function, line, log_buff);
             break;
         default:
             break;


platform/log/zephyr/src/mender-log.c
```diff
@@ -33,36 +33,47 @@ mender_log_init(void) {
     return MENDER_OK;
 }
 
+/* IMPORTANT: Please note that the default size of internal Zephyr logger buffer is 1024 bytes; if we need log buffer size 
+              to be more than that there is CONFIG_LOG_BUFFER_SIZE option allowing to extend the buffer size. */
+#define LOG_BUFFER_SIZE_BYTES (128)
```
lluiscampos, line 38: What are we trying to achieve by harcoding a number here at all, then.

The configurable value is from 128 to 65536, which means that for a user configuring it for the minimum, we won't be able to fit our logs. Either we have an explicit requirement of the logging buffer size of we _adapt_ our buffer size to `CONFIG_LOG_BUFFER_SIZE` minus the length for our prefix "(line": ".

platform/log/zephyr/src/mender-log.c
```diff
@@ -33,36 +33,47 @@ mender_log_init(void) {
...
 
-    /* Format message */
     va_list args;
     va_start(args, format);
-    vsnprintf(log, sizeof(log), format, args);
-    va_end(args);
+
+    char log_buff[LOG_BUFFER_SIZE_BYTES];
+    int  ret = vsnprintf(log_buff, LOG_BUFFER_SIZE_BYTES, format, args);
+
```
lluiscampos, line 51: I would keep `va_end(args);` here, is just more tidy.

platform/log/zephyr/src/mender-log.c
```diff
@@ -33,36 +33,47 @@ mender_log_init(void) {
...
+        LOG_ERR("logger error: log message formatting failed: %d", ret);
+        va_end(args);
+        return MENDER_FAIL;
+    }
 
     /* Switch depending log level */
     switch (level) {
         case MENDER_LOG_LEVEL_ERR:
-            LOG_ERR("%s (%d): %s", filename, line, log);
+            LOG_ERR("(%d): %s", line, log_buff);
```
lluiscampos, line 61: To be honest, having the line with no filename misses the point. To save buffer size we could trim the filename to its basename. Otherwise I vote for using the function and no filename/line.

lluiscampos (commented):
Thanks @pasinskim. See comments below.

Logging is currently disallowed in `main`. We already branched out an stable branch, so if you want we can enable it back in main for your experimenting.

pasinskim:
> Thanks @pasinskim. See comments below.
> 
> Logging is currently disallowed in `main`. We already branched out an stable branch, so if you want we can enable it back in main for your experimenting.

I think it is fine if it is off by default. We can provide instructions how to enable logging if needed.

platform/log/zephyr/src/mender-log.c
pasinskim: Not sure if I get it. We talking about two different buffers here. One is for printing messages (which is hard-coded now to `LOG_BUFFER_SIZE_BYTES (128)`). It means that if you have a log that is longer than 128 characters it will be truncated. 
The other buffer of size `CONFIG_LOG_BUFFER_SIZE` is for accumulating log messages and is Zephyr cyclic log buffer. The comment is mainly to make sure that the `LOG_BUFFER_SIZE_BYTES < CONFIG_LOG_BUFFER_SIZE`.

platform/log/zephyr/src/mender-log.c
pasinskim: Yes. The idea was to print log regardless of `va_end(args);`. If for whatever reason it fails we will still see the log. But maybe you are right it looks more tidy it it is next to `vsnprintf`.

platform/log/zephyr/src/mender-log.c
pasinskim: Makes sense. Let's get rid of line as well maybe. Doing some extra operation in logging, like getting basename prom the filename will make it even more heavy. I don't like already that we need to process arguments in the log function.

platform/log/zephyr/src/mender-log.c
lluiscampos: Okey, I didn't get that internal buffer was a circular buffer.

Nevermind, then.

platform/log/zephyr/src/mender-log.c
larsewi: We should have filename and line for at least the debug messages. Right?

platform/log/zephyr/src/mender-log.c
```diff
@@ -33,36 +33,47 @@ mender_log_init(void) {
...
 }
 
+/* IMPORTANT: Please note that the default size of internal Zephyr logger buffer is 1024 bytes; if we need log buffer size 
+              to be more than that there is CONFIG_LOG_BUFFER_SIZE option allowing to extend the buffer size. */
+#define LOG_BUFFER_SIZE_BYTES (128)
+
 mender_err_t
 mender_log_print(uint8_t level, const char *filename, const char *function, int line, char *format, ...) {
 
+    (void)filename;
```
vpodzime, line 43: Might as well use `ARG_UNUSED` in the argument list above instead.

platform/log/zephyr/src/mender-log.c
```diff
@@ -33,36 +33,47 @@ mender_log_init(void) {
...
-    va_end(args);
+
+    char log_buff[LOG_BUFFER_SIZE_BYTES];
+    int  ret = vsnprintf(log_buff, LOG_BUFFER_SIZE_BYTES, format, args);
+
+    if (ret < 0) {
+        LOG_ERR("logger error: log message formatting failed: %d", ret);
+        va_end(args);
+        return MENDER_FAIL;
+    }
```
vpodzime, line 56: Could we check `ret > LOG_BUFFER_SIZE_BYTES` and replace the end of the buffer with `(truncated)` or `...`?

platform/log/zephyr/src/mender-log.c
vpodzime: Yup! Those messages should have it.

platform/log/zephyr/src/mender-log.c
vpodzime, line 38: This is a maximum log message length, right? `LOG_BUFFER_SIZE` sounds to me like some memory where logs are stored before being flushed or something like that.

vpodzime (commented):
Looks good to me otherwise.

platform/log/zephyr/src/mender-log.c
lluiscampos: @vpodzime This is a miss-understanding between you and me, I figured it out only yesterday!

The Zephyr OS' `ARG_UNUSED` is not to be used in the argument list but here, while the one we defined in mender-utils.h is to be used in the argument list. Now, our custom one won't have effect for Zephyr OS builds and I also don't know how portable it is.

So yes, you were right [here by questioning me](https://github.com/mendersoftware/mender-mcu/pull/63#discussion_r1761000982), and I think we should stick to this style on platform code.

platform/log/zephyr/src/mender-log.c
lluiscampos: Yes, `CONFIG_LOG_BUFFER_SIZE` is a circular buffer for the logs while they wait to be consumed by the backend(s)

platform/log/zephyr/src/mender-log.c
lluiscampos: I think function name (and line number) is enough.

The issue with filename is that it resolves to something like `WEST_TOPDIR/modules/mender-mcu/core/src/mender-client.c (610)` - that is why I suggested to get the basename only. But this complicates the code, so I support removing it in favor of the function name and line.

platform/log/zephyr/src/mender-log.c
larsewi: Yes, but should be `>=`, because it returns number of characters that would have been printed, not including the trailing `'\0'`.

platform/log/zephyr/src/mender-log.c
lluiscampos: Of course! We should have started with `MENDER_` prefix from the beginning :melting_face: 

I leave it to you and Vratislav to decide if we need both for our code. Do you care @larsewi to follow-up on this on an independent PR?

vpodzime: > Of course! We should have started with `MENDER_` prefix from the beginning 🫠

I don't think that's necessary if we keep it in a header that's not included by any of the public API headers. But it doesn't hurt, I guess... Anyway, I prefer marking arguments as unused in the argument list rather than inside the functions.

platform/log/zephyr/src/mender-log.c
vpodzime: yup, it's a confusing name

platform/log/zephyr/src/mender-log.c
larsewi: >  Do you care @larsewi to follow-up on this on an independent PR?

On it

lluiscampos (approved):
It looks good :+1:

larsewi (approved):
Very nice 🚀

PR #63, opened 2024-09-13
chore: Guard definition of `ARG_UNUSED`

lluiscampos:
Fixes warning in Zephyr builds
```
/home/lluis/west/modules/mender-mcu/zephyr/../include/mender-utils.h:52: warning: "ARG_UNUSED" redefined
```

Previously defined in Zephyr toolchain:
* https://github.com/zephyrproject-rtos/zephyr/blob/v3.7.0/include/zephyr/toolchain/gcc.h#L263

### DIFF
--- a/include/mender-utils.h
+++ b/include/mender-utils.h
@@ -43,7 +43,9 @@ extern "C" {
 /**
  * @brief A utility macro to make marking unused arguments less noisy/disruptive
  */
+#ifndef ARG_UNUSED
 #define ARG_UNUSED __attribute__((unused))
+#endif
 
 /**
  * For variables only used in debug builds, in particular only in assert()


larsewi (approved):

include/mender-utils.h
```diff
@@ -43,7 +43,9 @@ extern "C" {
 /**
  * @brief A utility macro to make marking unused arguments less noisy/disruptive
  */
+#ifndef ARG_UNUSED
 #define ARG_UNUSED __attribute__((unused))
+#endif
```
vpodzime, line 48: Hmm, does it work? The definition there (below) doesn't look like something that would work the same way. At least not in an argument/parameter list.
```
#define ARG_UNUSED(x) (void)(x)```

vpodzime (commented):
ACK if it works. If not, we need a different name or undefine and define the macro to what we use it for.

include/mender-utils.h
lluiscampos: It works and it is broadly used in the kernel code.

PR #64, opened 2024-09-13
fix: Restore mender_save/get_update_state() for Zephyr

vpodzime:
Accidentally removed in de0fab1def36c91e65000727e4d7c251dd09a7b2.

### DIFF
--- a/platform/storage/zephyr/nvs/src/mender-storage.c
+++ b/platform/storage/zephyr/nvs/src/mender-storage.c
@@ -323,6 +323,67 @@ mender_storage_get_artifact_name(char **artifact_name) {
 }
 
 mender_err_t
+mender_storage_save_update_state(mender_update_state_t state, const char *artifact_type) {
+    assert(NULL != artifact_type);
+
+    size_t artifact_type_len;
+
+    if (!checked_nvs_write(&mender_storage_nvs_handle, MENDER_STORAGE_NVS_UPDATE_STATE, state, sizeof(state))) {
+        mender_log_error("Unable to save update state");
+        return MENDER_FAIL;
+    }
+
+    artifact_type_size = strlen(artifact_type) + 1;
+    if (!checked_nvs_write(&mender_storage_nvs_handle, MENDER_STORAGE_NVS_ARTIFACT_TYPE, artifact_type, artifact_type_size)) {
+        mender_log_error("Unable to save update state");
+        return MENDER_FAIL;
+    }
+
+    return MENDER_OK;
+}
+
+mender_err_t
+mender_storage_get_update_state(mender_update_state_t *state, char **artifact_type) {
+    assert(NULL != *artifact_type);
+
+    size_t       artifact_type_size;
+    ssize_t      n_read;
+    mender_err_t ret;
+
+    n_read = nvs_read(&mender_storage_nvs_handle, MENDER_STORAGE_NVS_UPDATE_STATE, NULL, 0);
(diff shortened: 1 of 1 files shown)


platform/storage/zephyr/nvs/src/mender-storage.c
```diff
@@ -322,6 +322,67 @@ mender_storage_get_artifact_name(char **artifact_name) {
...
+    assert(NULL != *artifact_type);
+
+    size_t       artifact_type_size;
+    ssize_t      n_read;
+    mender_err_t ret;
+
+    n_read = nvs_read(&mender_storage_nvs_handle, MENDER_STORAGE_NVS_UPDATE_STATE, NULL, 0);
+    if (0 == n_read) {
+        mender_log_debug("Update state record empty or unavailable");
+        return MENDER_NOT_FOUND;
```
lluiscampos, line 356: This is not correct, as the API can also return `-ENOENT` as not found. See https://github.com/mendersoftware/mender-mcu/commit/ff4d29e49bcaa0f928afb95c6e6578c809fafba1

platform/storage/zephyr/nvs/src/mender-storage.c
```diff
@@ -322,6 +322,67 @@ mender_storage_get_artifact_name(char **artifact_name) {
...
+
+mender_err_t
+mender_storage_get_update_state(mender_update_state_t *state, char **artifact_type) {
+    assert(NULL != *artifact_type);
+
+    size_t       artifact_type_size;
+    ssize_t      n_read;
+    mender_err_t ret;
+
+    n_read = nvs_read(&mender_storage_nvs_handle, MENDER_STORAGE_NVS_UPDATE_STATE, NULL, 0);
```
lluiscampos, line 353: And also here, for ESP flash driver, you will get invalid parameters always. See https://github.com/mendersoftware/mender-mcu/commit/10193f6e8069fb697c10fa74fbf639cfa5b2f734

lluiscampos (changes requested):
Both problems can be fixed by using the `nvs_read_alloc` wrapper that has the fixes/workarounds.

(I know that this PR is only restoring already approved code, but I missed that review)

vpodzime:
> Both problems can be fixed by using the `nvs_read_alloc` wrapper that has the fixes/workarounds.

Yeah, but we don't need to allocate space for the state, it's just an `enum`.

platform/storage/zephyr/nvs/src/mender-storage.c
```diff
@@ -322,6 +322,68 @@ mender_storage_get_artifact_name(char **artifact_name) {
...
+mender_storage_get_update_state(mender_update_state_t *state, char **artifact_type) {
+    assert(NULL != *artifact_type);
+
+    size_t        artifact_type_size;
+    ssize_t       n_read;
+    mender_err_t  ret;
+    unsigned char byte;
+
+    n_read = nvs_read(&mender_storage_nvs_handle, MENDER_STORAGE_NVS_UPDATE_STATE, &byte, 0);
+    if ((0 == n_read) || (-ENOENT == ret)) {
```
lluiscampos, line 355: here `ret` would need to be an integer, right? Being `mender_err_t` enum won't cast to `-ENOENT`

vpodzime: should check `n_read`, of course

lluiscampos (approved):

PR #65, opened 2024-09-16
fix: Fix Zephyr platform code for save/load state

lluiscampos:
By adding explicit pointer casts and amending a typo.

Additionally, add a `const` qualifier for the save method.

### DIFF
--- a/platform/storage/zephyr/nvs/src/mender-storage.c
+++ b/platform/storage/zephyr/nvs/src/mender-storage.c
@@ -344,9 +344,9 @@ mender_err_t
 mender_storage_save_update_state(mender_update_state_t state, const char *artifact_type) {
     assert(NULL != artifact_type);
 
-    size_t artifact_type_len;
+    size_t artifact_type_size;
 
-    if (!checked_nvs_write(&mender_storage_nvs_handle, MENDER_STORAGE_NVS_UPDATE_STATE, state, sizeof(state))) {
+    if (!checked_nvs_write(&mender_storage_nvs_handle, MENDER_STORAGE_NVS_UPDATE_STATE, (const void *)state, sizeof(state))) {
         mender_log_error("Unable to save update state");
         return MENDER_FAIL;
     }
@@ -385,7 +385,7 @@ mender_storage_get_update_state(mender_update_state_t *state, char **artifact_ty
     }
 
     *artifact_type = NULL;
-    ret            = nvs_read_alloc(&mender_storage_nvs_handle, MENDER_STORAGE_NVS_ARTIFACT_TYPE, artifact_type, &artifact_type_size);
+    ret            = nvs_read_alloc(&mender_storage_nvs_handle, MENDER_STORAGE_NVS_ARTIFACT_TYPE, (void **)artifact_type, &artifact_type_size);
     if (MENDER_OK != ret) {
         if (MENDER_NOT_FOUND == ret) {
             mender_log_error("Failed to read saved update state, ignoring");


include/mender-storage.h
```diff
@@ -81,7 +81,7 @@ mender_err_t mender_storage_get_deployment_data(char **deployment_data);
  * @param artifact_type artifact type of the update module we are saving the state of
  * @param state state of the update module
  */
-mender_err_t mender_storage_save_update_state(mender_update_state_t state, const char *artifact_type);
+mender_err_t mender_storage_save_update_state(const mender_update_state_t state, const char *artifact_type);
```
vpodzime, line 84: This has no effect here in the function declaration.

platform/storage/zephyr/nvs/src/mender-storage.c
```diff
@@ -341,18 +341,18 @@ mender_storage_get_artifact_name(char **artifact_name) {
...
 
 mender_err_t
-mender_storage_save_update_state(mender_update_state_t state, const char *artifact_type) {
+mender_storage_save_update_state(const mender_update_state_t state, const char *artifact_type) {
     assert(NULL != artifact_type);
 
     size_t artifact_type_len;
 
-    if (!checked_nvs_write(&mender_storage_nvs_handle, MENDER_STORAGE_NVS_UPDATE_STATE, state, sizeof(state))) {
+    if (!checked_nvs_write(&mender_storage_nvs_handle, MENDER_STORAGE_NVS_UPDATE_STATE, (const void *)state, sizeof(state))) {
```
vpodzime, line 349: Should it be `void *` or `unsigned char *`?

platform/storage/zephyr/nvs/src/mender-storage.c
```diff
@@ -341,18 +341,18 @@ mender_storage_get_artifact_name(char **artifact_name) {
...
         mender_log_error("Unable to save update state");
         return MENDER_FAIL;
     }
 
-    artifact_type_size = strlen(artifact_type) + 1;
-    if (!checked_nvs_write(&mender_storage_nvs_handle, MENDER_STORAGE_NVS_ARTIFACT_TYPE, artifact_type, artifact_type_size)) {
+    artifact_type_len = strlen(artifact_type) + 1;
+    if (!checked_nvs_write(&mender_storage_nvs_handle, MENDER_STORAGE_NVS_ARTIFACT_TYPE, artifact_type, artifact_type_len)) {
```
vpodzime, line 355: It's not a length with the `+1`, that's why I preferred the word *size*

vpodzime (commented):
Thanks!

include/mender-storage.h
lluiscampos: Do you mean that it is passed by copy? I'll remove it as this is not really what the PR was about but I don't understand what you mean :man_student:

platform/storage/zephyr/nvs/src/mender-storage.c
lluiscampos: The `nvs_write` API call takes `void *`. We could cast to char, but then it will anyway be interpreted as void... so I think is best to do it at once here.

platform/storage/zephyr/nvs/src/mender-storage.c
lluiscampos: :ok_hand: - I will modify the declaration instead, then.

include/mender-storage.h
larsewi: What he means is that the `const` keyword (in this case) has no effect in the function declaration. If you put it in the function definition, it will have an effect.

larsewi (approved):

include/mender-storage.h
lluiscampos: Right. But it was both in the declaration and in the definition (both for Zephyr and POSIX). I did revert this, anyway.

PR #70, opened 2024-09-16
fix: Fix (more) platform code for update state storage

lluiscampos:
We need to cast to `void * ` the _address_ of `state`.

### DIFF
--- a/platform/storage/zephyr/nvs/src/mender-storage.c
+++ b/platform/storage/zephyr/nvs/src/mender-storage.c
@@ -346,7 +346,7 @@ mender_storage_save_update_state(mender_update_state_t state, const char *artifa
 
     size_t artifact_type_size;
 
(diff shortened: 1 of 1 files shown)


larsewi (approved):

platform/storage/zephyr/nvs/src/mender-storage.c
```diff
@@ -346,7 +346,7 @@ mender_storage_save_update_state(mender_update_state_t state, const char *artifa
 
     size_t artifact_type_size;
 
-    if (!checked_nvs_write(&mender_storage_nvs_handle, MENDER_STORAGE_NVS_UPDATE_STATE, (const void *)state, sizeof(state))) {
+    if (!checked_nvs_write(&mender_storage_nvs_handle, MENDER_STORAGE_NVS_UPDATE_STATE, (const void *)&state, sizeof(state))) {
```
vpodzime, line 349: I think you don't need the `void*` cast anymore, the missing `&` was the real issue here.

lluiscampos: Thanks! That makes sense

PR #71, opened 2024-09-16
Update `feature-update_modules` from latest `main`

lluiscampos:
The main conflicts were around the state machine and the scheduler
    "rewrite". I solved them by preferring the alternative scheduler,
    removing the semaphore and mutexes, while keeping the update modules
    state machine.

### DIFF
--- a/core/src/mender-client.c
+++ b/core/src/mender-client.c
@@ -104,22 +104,15 @@ static const char *update_state_str[N_MENDER_UPDATE_STATES] = {
 #endif
 
 /**
- * @brief Counter and mutex for the management of network connect/release callbacks
- */
-static uint8_t mender_client_network_count = 0;
-static void   *mender_client_network_mutex = NULL;
-
-/**
  * @brief Deployment data (ID, artifact name and payload types), used to report deployment status after rebooting
  */
 static cJSON *mender_client_deployment_data = NULL;
 
 /**
- * @brief Mender client update modules list and mutex
+ * @brief Mender client update modules list
  */
 static mender_update_module_t **mender_update_modules_list  = NULL;
 static size_t                   mender_update_modules_count = 0;
-static void                    *mender_update_modules_mutex = NULL;
 
 /**
  * @brief Update module being used by the current deployment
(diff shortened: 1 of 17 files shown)


core/src/mender-client.c
```diff
@@ -386,117 +349,49 @@ mender_client_activate(void) {
...
-mender_client_deactivate(void) {
-
-    mender_err_t ret;
-
-#ifdef CONFIG_MENDER_CLIENT_INVENTORY
-    if (MENDER_OK != (ret = mender_inventory_deactivate())) {
-        mender_log_error("Unable to deactivate the inventory functionality");
-        /* keep going on, we want to do as much cleanup as possible */
-    }
-#endif /* CONFIG_MENDER_CLIENT_INVENTORY */
```
vpodzime, line 404: `mender_inventory_activate()` is kept in place above. I think we should keep the deactivation here too. We require no inventory in builds so this should not be a functional problem. I don't see a reason to throw this code away.

platform/scheduler/zephyr/src/mender-scheduler.c
```diff
@@ -17,10 +18,68 @@
  * limitations under the License.
  */
 
+#ifdef CONFIG_MENDER_CLIENT_ADD_ON_INVENTORY
```
vpodzime, line 21: It's no longer an addon.

vpodzime (commented):
Looks good to me otherwise.

larsewi (approved):
The parts of the code that I have touched looks good :)

core/src/mender-client.c
lluiscampos: Right. Thank you.

lluiscampos: Okey. But the issue is that the caller `mender_client_deactivate` is not kept. I could keep `mender_client_deactivate` but calling it will be undefined behavior.

I think is best to remove the whole thing as we need to re-implement with the new scheduler on [MEN-7536](https://northerntech.atlassian.net/browse/MEN-7536)

PR #76, opened 2024-09-19
fix: store state in deployment data

danielskinstad:
Removed the key and corresponding functions for storing the update_state
and the artifact type. It is now a part of the deployment data.

Changelog: Title
Ticket: MEN-7515

### DIFF
--- a/core/src/mender-client.c
+++ b/core/src/mender-client.c
@@ -868,9 +868,7 @@ mender_client_check_deployment(mender_api_deployment_data_t **deployment_data) {
         mender_log_warning("Unexpected stale deployment data");
         mender_delete_deployment_data(mender_client_deployment_data);
     }
-    if (MENDER_OK
-        != (mender_create_deployment_data(
-            deployment->id, deployment->artifact_name, NULL /* We will populate later */, NULL /* TODO: MEN-7515 */, &mender_client_deployment_data))) {
+    if (MENDER_OK != (mender_create_deployment_data(deployment->id, deployment->artifact_name, &mender_client_deployment_data))) {
         /* Error already logged */
         return MENDER_FAIL;
     }
--- a/core/src/mender-client.c
+++ b/core/src/mender-client.c
@@ -518,7 +518,6 @@ REBOOT:
 
     /* Delete pending deployment */
     mender_storage_delete_deployment_data();
-    mender_storage_delete_update_state();
 
     /* Invoke restart callback, application is responsible to shutdown properly and restart the system */
     if (NULL != mender_client_callbacks.restart) {
@@ -877,6 +876,27 @@ mender_client_check_deployment(mender_api_deployment_data_t **deployment_data) {
 }
 
 static mender_err_t
+set_and_store_state(const mender_update_state_t state) {
+
+    /*
+     * Set the state in `mender_client_deployment_data` and write it to the nvs
+     */
+
+    /* Set state in deployment data */
+    if (MENDER_OK != mender_deployment_data_set_state(mender_client_deployment_data, state)) {
+        mender_log_error("Failed to set deployment data state");
+        return MENDER_FAIL;
+    }
+
+    /* Store deployment data */
+    if (MENDER_OK != mender_set_deployment_data(mender_client_deployment_data)) {
+        mender_log_error("Failed to store deployment data");
+        return MENDER_FAIL;
+    }
+    return MENDER_OK;
+}
+
+static mender_err_t
 mender_client_update_work_function(void) {
     mender_err_t ret = MENDER_OK;
 
@@ -896,9 +916,10 @@ mender_client_update_work_function(void) {
     }
 
     {
-        char                 *artifact_type;
+        const char           *artifact_type;
         mender_update_state_t update_state_resume;
-        if (MENDER_OK == (ret = mender_storage_get_update_state(&update_state_resume, &artifact_type))) {
+        if (MENDER_OK == (ret = mender_deployment_data_get_state(mender_client_deployment_data, &update_state_resume))
+            && MENDER_OK == mender_deployment_data_get_payload_type(mender_client_deployment_data, &artifact_type)) {
             update_state = update_state_resume;
             mender_log_debug("Resuming from state %s", update_state_str[update_state]);
             mender_update_module = mender_client_get_update_module(artifact_type);
@@ -907,11 +928,8 @@ mender_client_update_work_function(void) {
                 mender_log_error("No update module found for artifact type '%s'", artifact_type);
                 mender_client_publish_deployment_status(deployment_id, MENDER_DEPLOYMENT_STATUS_FAILURE);
                 mender_storage_delete_deployment_data();
-                mender_storage_delete_update_state();
-                free(artifact_type);
                 goto END;
             }
-            free(artifact_type);
         }
     }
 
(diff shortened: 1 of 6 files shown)


core/src/mender-deployment-data.c
```diff
@@ -239,6 +239,9 @@ mender_create_deployment_data(const char *id, const char *artifact_name, const c
 
 mender_err_t
 __mender_deployment_data_get_string(const mender_deployment_data_t *deployment_data, const char *key, const char **str) {
+    if (NULL == deployment_data) {
+        return MENDER_NOT_FOUND;
+    }
 
     assert(NULL != deployment_data);
```
larsewi, line 246: This assert is not needed anymore

core/src/mender-deployment-data.c
```diff
@@ -279,6 +282,41 @@ __mender_deployment_data_set_string(mender_deployment_data_t *deployment_data, c
...
+    assert(NULL != key);
+    assert(NULL != state);
+
+    cJSON *item;
+    if (NULL == (item = cJSON_GetObjectItemCaseSensitive(deployment_data, key))) {
+        return MENDER_FAIL;
+    }
+
+    /* Assign state value to state pointer */
+    double state_value = cJSON_GetNumberValue(item);
```
larsewi, line 301: What happens if `item` is JSON null?

include/mender-deployment-data.h
```diff
@@ -121,12 +132,17 @@ mender_err_t __mender_deployment_data_get_string(const mender_deployment_data_t
...
+
+/**
+ * @brief Get state
  * @param deployment_data Deployment data
- * @param name State name
+ * @param State state
  * @return MENDER_OK on success, otherwise MENDER_FAIL
  */
-#define mender_deployment_data_get_state(deployment_data, name) __mender_deployment_data_get_string(deployment_data, MENDER_DEPLOYMENT_DATA_KEY_STATE, state)
+#define mender_deployment_data_get_state(deployment_data, state) __mender_deployment_data_get_state(deployment_data, MENDER_DEPLOYMENT_DATA_KEY_STATE, state);
```
larsewi, line 145: Since this is a "function" with it's own definition, we don't really need to use this macro. And we know that the key is always `MENDER_DEPLOYMENT_DATA_KEY_STATE`, so we don't need to pass it as an argument.

core/src/mender-client.c
```diff
@@ -926,21 +930,21 @@ mender_client_update_work_function(void) {
...
-            mender_storage_save_update_state(update_state, mender_update_module->artifact_type); \
-        }                                                                                        \
-        ret = MENDER_OK;                                                                         \
-        continue;                                                                                \
+#define NEXT_STATE                                                                         \
+    if (MENDER_OK == ret) {                                                                \
+        update_state = update_state_transitions[update_state].success;                     \
+        assert(NULL != mender_update_module);                                              \
+        mender_log_debug("Entering state %s", update_state_str[update_state]);             \
+        mender_deployment_data_set_state(mender_client_deployment_data, update_state);     \
```
larsewi, line 938: The previous one used to write to non-volatile storage. This one only writes to volatile memory. Is this taken care of?

danielskinstad: I believe so, it's stored in `mender_client_deployment_data` which is written to the store on `MENDER_UPDATE_STATE_REBOOT`

core/src/mender-deployment-data.c
danielskinstad: That is checked in the if-statement above

include/mender-deployment-data.h
```diff
@@ -35,6 +36,8 @@ extern "C" {
 #define MENDER_DEPLOYMENT_DATA_KEY_STATE                  "state"
 #define MENDER_DEPLOYMENT_DATA_KEY_STATE_DATA_STORE_COUNT "state_data_store_count"
 
+#define MENDER_DEPLOYMENT_DATA_KEY_STATE "state"
```
vpodzime, line 39: It's defined just 3 lines above.

core/src/mender-deployment-data.c
```diff
@@ -279,6 +282,41 @@ __mender_deployment_data_set_string(mender_deployment_data_t *deployment_data, c
     return MENDER_OK;
 }
 
+mender_err_t
+__mender_deployment_data_get_state(mender_deployment_data_t *deployment_data, const char *key, mender_update_state_t *state) {
```
vpodzime, line 286: Why do we need `key`? Are we ever going to have multiple states under different keys?

include/mender-deployment-data.h
lluiscampos, line 39: Duplicated

include/mender-deployment-data.h
```diff
@@ -78,7 +81,7 @@ mender_err_t mender_get_deployment_data(mender_deployment_data_t **deployment_da
  * @return MENDER_OK on success, otherwise MENDER_FAIL
  */
 mender_err_t mender_create_deployment_data(
-    const char *id, const char *artifact_name, const char *provides, const char *name, mender_deployment_data_t **deployment_data);
+    const char *id, const char *artifact_name, const char *provides, const char *state, mender_deployment_data_t **deployment_data);
```
lluiscampos, line 84: Pre-existing,

Because we are actually passing `NULL`, it "works". But this should instead be
```suggestion
    const char *id, const char *artifact_name, const char *provides, const mender_update_state_t *state, mender_deployment_data_t **deployment_data);
```

What is more, if we are only passing NULL, shouldn't we just remove it from the arguments? (same for provides) Or is it value in leaving it for some reason that I cannot see?

include/mender-deployment-data.h
```diff
@@ -89,6 +92,14 @@ mender_err_t mender_create_deployment_data(
  */
 mender_err_t mender_deployment_data_add_payload_type(mender_deployment_data_t *deployment_data, const char *payload_type);
 
+/**
+ * @brief Get payload type
+ * @param deployment_data Deployment data
+ * @param payload_type Payload type
+ * @return MENDER_OK on success, otherwise MENDER_FAIL
```
lluiscampos, line 99: I would add here a `@note` -> Because we don't support Artifacts with more than one payload, it will return always the first payload type from the list.

core/src/mender-client.c
danielskinstad: Right, so should we just write the deployment data each time we save the state? Right now the deployment data is only written on `MENDER_UPDATE_STATE_REBOOT`, but should we write it each time the state is set?

core/src/mender-deployment-data.c
danielskinstad: We don't need it, residue from the definition of `__mender_deployment_data_set_string`

include/mender-deployment-data.h
danielskinstad: I believe we can remove them from the arguments, unless you added them there for a specific reason @larsewi

core/src/mender-client.c
```diff
@@ -518,7 +518,6 @@ mender_client_initialization_work_function(void) {
 
     /* Delete pending deployment */
     mender_storage_delete_deployment_data();
```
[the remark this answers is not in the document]
danielskinstad: Right, you think I should rename everything that's named `deployment data` to `update data` (including mender-deployment-data.c etc.?)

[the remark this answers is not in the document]
lluiscampos: > We discussed should call the data update data (...)

To be fair with Daniel in the ticket we wrote "consider renaming..."

lluiscampos (approved):
It looks good to me - it seems like all my concerns were addressed. Thanks!

include/mender-deployment-data.h
larsewi: Yes, no issue in removing them ;) But should add a comment in the doc string that these must be populated later

include/mender-deployment-data.h
```diff
@@ -118,12 +128,12 @@ mender_err_t __mender_deployment_data_get_string(const mender_deployment_data_t
 #define mender_deployment_data_get_id(deployment_data, id) __mender_deployment_data_get_string(deployment_data, MENDER_DEPLOYMENT_DATA_KEY_ID, id)
 
 /**
- * @brief Get state name
+ * @brief Get state
  * @param deployment_data Deployment data
- * @param name State name
+ * @param State state
```
larsewi, line 133: The param should be lowercase `state` not `State`

larsewi (approved):

core/src/mender-client.c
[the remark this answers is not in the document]
vpodzime: Yeah, I but IIRC we discussed it on Slack or somewhere else too...

core/src/mender-client.c
```diff
@@ -1027,20 +1045,15 @@ mender_client_update_work_function(void) {
...
+                 * wrong. */
+                set_and_store_state(MENDER_UPDATE_STATE_VERIFY_REBOOT);
                 if ((MENDER_OK == ret) && (NULL != mender_update_module->callbacks[update_state])) {
-                    /* Save the next state before running the reboot callback --
-                     * if there is an interrupt (power, crash,...) right after,
-                     * it will reboot anyway so after the new boot, reboot
-                     * verification should happen anyway, the callback in that
-                     * state should be able to see if things went well or
-                     * wrong. */
-                    mender_storage_save_update_state(MENDER_UPDATE_STATE_VERIFY_REBOOT, mender_update_module->artifact_type);
```
vpodzime, line 1043: I think we still want do do `set_and_store_state()` here iff there is a reboot callback. If there isn't, it will be called by `NEXT_STATE` below anyway. It just wouldn't be called twice in such a case.

vpodzime (commented):
Looks good to me otherwise.

core/src/mender-client.c
danielskinstad: I moved it outside of the if-statement, but I moved it back now, because it's stored in `NEXT_STATE` as you said

core/src/mender-client.c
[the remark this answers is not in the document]
danielskinstad: I can do it in a separate PR

PR #84, opened 2024-09-27
Explicit error for compressed Artifacts

danielskinstad:


core/src/mender-artifact.c
```diff
@@ -179,6 +179,24 @@ mender_artifact_get_ctx(mender_artifact_ctx_t **ctx) {
...
+static bool
+is_compressed(const char *filename) {
+
+    /* Return true if artifact is compressed - otherwise false
+     */
+
+    const char  *compression_suffixes[]    = { ".gz", ".xz", ".zst" };
+    const size_t compression_suffixes_size = 3;
+
+    for (size_t i = 0; i < compression_suffixes_size; i++) {
```
larsewi, line 191: Consider:
- making it `static`. This way it does not get initialized every time you call the function
- making it `NULL`-terminated. This way you do not have to hard code or compute the size
- using `int`. It's usually the size that the CPU is optimized for
```suggestion
    static const char  *compression_suffixes[]    = { ".gz", ".xz", ".zst", NULL };

    for (int i = 0; NULL != compression_suffixes[i]; i++) {
```

larsewi (approved):
🚀

core/src/mender-artifact.c
```diff
@@ -179,6 +179,23 @@ mender_artifact_get_ctx(mender_artifact_ctx_t **ctx) {
     return MENDER_OK;
 }
 
+static bool
+is_compressed(const char *filename) {
+
+    /* Return true if artifact is compressed - otherwise false
+     */
```
vpodzime, line 186: I don't think we need this important comment :sweat_smile:

core/src/mender-artifact.c
```diff
@@ -204,6 +221,12 @@ mender_artifact_process_data(mender_artifact_ctx_t *ctx,
     /* Parse data */
     do {
 
+        /* We do not support compressed artifacts */
+        if (is_compressed(ctx->file.name)) {
```
vpodzime, line 225: Is this derived from the URL?

core/src/mender-client.c
```diff
@@ -1170,15 +1170,18 @@ mender_client_download_artifact_callback(char *type, cJSON *meta_data, char *fil
...
-    static size_t download_progress = 0;
-    /* New update */
-    if (0 == index) {
-        download_progress = 0;
-    }
-    /* Update every 10% */
-    if (((index * 10) / size) > download_progress) {
-        download_progress = (index * 10) / size;
-        mender_log_info("Downloading '%s' %zu0%%... [%zu/%zu]", type, download_progress, index, size);
+    if (0 < size) {
```
vpodzime, line 1173: I'd do `size > 0`. This is not a comparison that could be confused with an assignment, no need to have the constant on the left. But YMMV.

vpodzime (commented):
Looks good to me otherwise.

core/src/mender-client.c
```diff
@@ -1170,15 +1170,18 @@ mender_client_download_artifact_callback(char *type, cJSON *meta_data, char *fil
...
-    static size_t download_progress = 0;
-    /* New update */
-    if (0 == index) {
-        download_progress = 0;
-    }
-    /* Update every 10% */
-    if (((index * 10) / size) > download_progress) {
-        download_progress = (index * 10) / size;
-        mender_log_info("Downloading '%s' %zu0%%... [%zu/%zu]", type, download_progress, index, size);
+    if (size > 0) {
```
lluiscampos, line 1173: Thanks

lluiscampos (approved):

PR #85, opened 2024-09-30
feat: define Device Type at build time

larsewi:
You can now define Device Type at build time using Kconfig options. Kconfig options are ignored if device type is also specified in `mender_client_config_t`.

Ticket [MEN-7547](https://northerntech.atlassian.net/browse/MEN-7547) says to remove `device_type` from `mender_client_config_t`. But I don't see the reason why we cannot have both.

[MEN-7547]: https://northerntech.atlassian.net/browse/MEN-7547?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/src/mender-client.c
```diff
@@ -244,13 +251,17 @@ mender_err_t
...
 
-    mender_client_config.device_type = config->device_type;
-
+    if ((NULL != config->device_type) && (0 != strlen(config->device_type))) {
+        /* Use client config unless string is NULL or empty */
+        mender_client_config.device_type = config->device_type;
+    } else {
+        /* Otherwise, use Kconfig */
+        mender_client_config.device_type = CONFIG_MENDER_DEVICE_TYPE;
+    }
```
lluiscampos, line 264: `CONFIG_MENDER_DEVICE_TYPE` can be null, so:
```suggestion
    } else if  (NULL != CONFIG_MENDER_DEVICE_TYPE ){
        /* Otherwise, use Kconfig */
        mender_client_config.device_type = CONFIG_MENDER_DEVICE_TYPE;
    } else {
       /* Error... */
    }
```

CMakeLists.txt
```diff
@@ -40,6 +40,11 @@ if (NOT CONFIG_MENDER_SERVER_TENANT_TOKEN)
 else()
     message(STATUS "Using custom '${CONFIG_MENDER_SERVER_TENANT_TOKEN}' server tenant token")
 endif()
+if (NOT CONFIG_MENDER_DEVICE_TYPE)
+    message(STATUS "Using default device type")
+else()
+    message(STATUS "Using custom '${CONFIG_MENDER_DEVICE_TYPE}' device type")
+endif()
```
lluiscampos, line 47: For yours and other readers information,
The main `CMakeLists.txt` file is ignored when building the project in Zephyr mode.

core/src/mender-client.c
```diff
@@ -244,13 +251,17 @@ mender_err_t
...
 
     assert(NULL != config);
-    assert(NULL != config->device_type);
     assert(NULL != callbacks);
     assert(NULL != callbacks->restart);
     mender_err_t ret;
 
-    mender_client_config.device_type = config->device_type;
-
+    if ((NULL != config->device_type) && (0 != strlen(config->device_type))) {
```
lluiscampos, line 258: It is a good call to leave the device type, which aligns the API with the server url, for example

core/src/mender-client.c
lluiscampos: Or to have an explicit follow-up check, as it is done for `mender_client_config.host` :point_down:

lluiscampos (approved):

zephyr/Kconfig
```diff
@@ -46,6 +46,12 @@ if MENDER_MCU_CLIENT
             help
                 Set the Mender server Tenant Token, to be used with https://hosted.mender.io. Retrieve it from the "Organization and billing" settings of your account.
 
+        config MENDER_DEVICE_TYPE
+            string "Mender Device Type"
+            default BOARD_TARGET
```
lluiscampos, line 51: Have you tried this? The idea is not to be literally `BOARD_TARGET` but to use the already defined variable from Zephyr KConfig

core/src/mender-client.c
vpodzime: Leave the device type?

vpodzime, line 258: Let's add a macro `EMPTY_STRING()` checking for `!= NULL` and then checking `[0] == '\0'`. It's quite a bit cheaper then calling `strlen()`.

zephyr/Kconfig
vpodzime: Yeah, I guess we should do 
```
#ifndef CONFIG_MENDER_DEVICE_TYPE
#define CONFIG_MENDER_DEVICE_TYPE CONFIG_BOARD_TARGET
```

vpodzime (commented):
Looks good to me otherwise.

core/src/mender-client.c
lluiscampos: I was referring to his point in the PR description:

> Ticket [MEN-7547](https://northerntech.atlassian.net/browse/MEN-7547) says to remove device_type from mender_client_config_t. But I don't see the reason why we cannot have both.

CMakeLists.txt
larsewi: Should I not touch this file?

zephyr/Kconfig
larsewi: Yes, I tried it and it did work just fine. It get's the value from the Zephyr Kconfig

larsewi: Contents in `west build -t menuconfig` after a pristine build:
Screenshot 2024-10-02 at 16 44 31

lluiscampos: Fantastic. Thank you for double-checking

CMakeLists.txt
lluiscampos: We can touch it as much as we want to make our POSIX build nice, but I am just saying that is ignored on the integration builds.

zephyr/Kconfig
vpodzime: Nice! :heart_eyes:

vpodzime (approved):
ACK

PR #91, opened 2024-10-04
feat: implement ArtifactRollbackReboot retry

danielskinstad:
If `ArtifactRollbackVerifyReboot` fails, `ArtifactRollbackReboot` will be retried `CONFIG_MENDER_ROLLBACK_REBOOT_LOOP_RETRIES` times.

Ticket: MEN-7496
Changelog: Title

### DIFF
--- a/core/src/mender-client.c
+++ b/core/src/mender-client.c
@@ -96,8 +96,8 @@ static const struct mender_update_state_transition_s update_state_transitions[N_
     /* MENDER_UPDATE_STATE_COMMIT                 */ { MENDER_UPDATE_STATE_CLEANUP, MENDER_UPDATE_STATE_ROLLBACK },
     /* MENDER_UPDATE_STATE_CLEANUP                */ { MENDER_UPDATE_STATE_END, MENDER_UPDATE_STATE_END },
     /* MENDER_UPDATE_STATE_ROLLBACK               */ { MENDER_UPDATE_STATE_ROLLBACK_REBOOT, MENDER_UPDATE_STATE_FAILURE },
-    /* MENDER_UPDATE_STATE_ROLLBACK_REBOOT        */ { MENDER_UPDATE_STATE_ROLLBACK_VERIFY_REBOOT, MENDER_UPDATE_STATE_FAILURE },
-    /* MENDER_UPDATE_STATE_ROLLBACK_VERIFY_REBOOT */ { MENDER_UPDATE_STATE_FAILURE, MENDER_UPDATE_STATE_FAILURE },
+    /* MENDER_UPDATE_STATE_ROLLBACK_REBOOT        */ { MENDER_UPDATE_STATE_ROLLBACK_VERIFY_REBOOT, MENDER_UPDATE_STATE_ROLLBACK_VERIFY_REBOOT },
+    /* MENDER_UPDATE_STATE_ROLLBACK_VERIFY_REBOOT */ { MENDER_UPDATE_STATE_FAILURE, MENDER_UPDATE_STATE_ROLLBACK_REBOOT },
     /* MENDER_UPDATE_STATE_FAILURE                */ { MENDER_UPDATE_STATE_CLEANUP, MENDER_UPDATE_STATE_CLEANUP },
 };
 
@@ -1149,6 +1149,19 @@ mender_client_update_work_function(void) {
                 if (NULL != mender_update_module->callbacks[update_state]) {
                     ret = mender_update_module->callbacks[update_state](update_state, (mender_update_state_data_t)NULL);
                 }
+
+                if (MENDER_OK != ret) {
+                    /* If the rollback verify reboot fails,
+                     * we will retry the rollback reboot.
+                     *
+                     * TODO: fix this comment / document infinite
+                     * loop detection once MEN-7516 is implemented.
+                     * As it is now, the rollback will be retried
+                     * infinitely, or until it succeeds
+                     */
+                    mender_log_error("Rollback verify reboot failed. Retry rollback reboot");
+                }
+
                 NEXT_STATE;
                 /* fallthrough */
 


core/src/mender-client.c
```diff
@@ -1143,6 +1150,25 @@ mender_client_update_work_function(void) {
...
+                    size_t retry_rollback_reboot_counter;
+                    mender_deployment_data_retry_rollback_reboot_counter(mender_client_deployment_data, &retry_rollback_reboot_counter);
+                    if (retry_rollback_reboot_counter <= CONFIG_MENDER_ROLLBACK_REBOOT_LOOP_RETRIES) {
+                        mender_log_info("Rollback verify reboot failed. Retry rollback reboot %zu/%d",
+                                        retry_rollback_reboot_counter,
+                                        CONFIG_MENDER_ROLLBACK_REBOOT_LOOP_RETRIES);
+                        update_state = MENDER_UPDATE_STATE_ROLLBACK_REBOOT;
+                        mender_log_debug("Entering state %s", update_state_str[update_state]);
+                        set_and_store_state(update_state);
+                        continue;
```
lluiscampos, line 1168: We will need some wait between retries. I don't think we have anything else in the code _waiting_, though so I don't know how to handle it.

core/src/mender-deployment-data.c
```diff
@@ -59,6 +59,7 @@ validate_deployment_data(const cJSON *deployment_data) {
         { .key = MENDER_DEPLOYMENT_DATA_KEY_PROVIDES, .type = cJSON_IsString },               /* Artifact provides (filtered on clears provides) */
         /* { .key = MENDER_DEPLOYMENT_DATA_KEY_STATE, .type = cJSON_IsString }, */            /* TODO: MEN-7515: State name */
         { .key = MENDER_DEPLOYMENT_DATA_KEY_STATE_DATA_STORE_COUNT, .type = cJSON_IsNumber }, /* State data store count */
+        { .key = MENDER_DEPLOYMENT_DATA_KEY_ROLLBACK_REBOOT_COUNT, .type = cJSON_IsNumber },  /* Rollback reboot count */
```
lluiscampos, line 62: I don't think we need to save this counter into the non-volatile storage.

It will retry up to the configured times during the current boot, and on spontaneous reboots it will retry from scratch (which might be suboptimal) but there is no risk of it getting stuck forever because we have the overall "State data store count".

Is there an extra benefit that am I missing? My motivation for not adding it is to keep the state data as simple as possible (it is already too complex IMO).

danielskinstad: > Is there an extra benefit that am I missing?
I don't think so. I didn't think of the state data store count, so I thought it could get stuck. 
After looking at the code again, I agree that we don't need to store the counter

core/src/mender-client.c
danielskinstad: We have a `k_sleep` in mender-net.c, should I add a sleep? Do you have a suggestion or an interval in mind?

lluiscampos: As discussed in person,

I was wrong here. This is looping through two separate states, therefore we don't need to sleep between them...

core/src/mender-deployment-data.c
lluiscampos: Discussed in person, we can drop the counter altogether.

danielskinstad:
Note that this might cause an infinite loop with the code in its current stage, however there is already a ticket for this (MEN-7516). The state data counter will give an error if it's exceeded, but we don't check what `set_and_store_state` returns atm

core/src/mender-client.c
```diff
@@ -1143,6 +1143,19 @@ mender_client_update_work_function(void) {
...
                     ret = mender_update_module->callbacks[update_state](update_state, (mender_update_state_data_t)NULL);
                 }
+
+                if (MENDER_OK != ret) {
+                    /* If the rollback verify reboot fails, we will retry the
+                     * rollback reboot.
+                     * Will be automatically aborted when maximum state data count is reached
+                     */
+                    mender_log_info("Rollback verify reboot failed. Retry rollback reboot");
+                    update_state = MENDER_UPDATE_STATE_ROLLBACK_REBOOT;
```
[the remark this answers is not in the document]
danielskinstad: Good question :sweat_smile: Left over from a previous implementation

core/src/mender-client.c
```diff
@@ -1143,6 +1143,19 @@ mender_client_update_work_function(void) {
                 if (NULL != mender_update_module->callbacks[update_state]) {
                     ret = mender_update_module->callbacks[update_state](update_state, (mender_update_state_data_t)NULL);
                 }
+
+                if (MENDER_OK != ret) {
+                    /* If the rollback verify reboot fails, we will retry the
+                     * rollback reboot.
+                     * Will be automatically aborted when maximum state data count is reached
```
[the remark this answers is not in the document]
danielskinstad: Well, the comment might be a bit misleading/incorrect. But we are implementing an infinite loop detection, which I guess will be based on the state data count? I can modify / remove the comment, as there's nothing to detect an infinite loop atm anyways

core/src/mender-client.c
```diff
@@ -96,8 +96,8 @@ static const struct mender_update_state_transition_s update_state_transitions[N_
     /* MENDER_UPDATE_STATE_COMMIT                 */ { MENDER_UPDATE_STATE_CLEANUP, MENDER_UPDATE_STATE_ROLLBACK },
     /* MENDER_UPDATE_STATE_CLEANUP                */ { MENDER_UPDATE_STATE_END, MENDER_UPDATE_STATE_END },
     /* MENDER_UPDATE_STATE_ROLLBACK               */ { MENDER_UPDATE_STATE_ROLLBACK_REBOOT, MENDER_UPDATE_STATE_FAILURE },
-    /* MENDER_UPDATE_STATE_ROLLBACK_REBOOT        */ { MENDER_UPDATE_STATE_ROLLBACK_VERIFY_REBOOT, MENDER_UPDATE_STATE_FAILURE },
-    /* MENDER_UPDATE_STATE_ROLLBACK_VERIFY_REBOOT */ { MENDER_UPDATE_STATE_FAILURE, MENDER_UPDATE_STATE_FAILURE },
+    /* MENDER_UPDATE_STATE_ROLLBACK_REBOOT        */ { MENDER_UPDATE_STATE_ROLLBACK_VERIFY_REBOOT, MENDER_UPDATE_STATE_ROLLBACK_VERIFY_REBOOT },
+    /* MENDER_UPDATE_STATE_ROLLBACK_VERIFY_REBOOT */ { MENDER_UPDATE_STATE_FAILURE, MENDER_UPDATE_STATE_ROLLBACK_VERIFY_REBOOT },
```
vpodzime, line 100: So when a rollback reboot fails, we enter rollback reboot verification which is basically pre-determined to fail because the rollback reboot itself failed. I don't see how that's supposed to work.

vpodzime (commented):
Plus I still think this "rollback reboot retry" behavior should be configurable

core/src/mender-client.c
danielskinstad: In the c++ client there is no failure state for rollback reboot, it goes to verification or state loop detection. I guess the idea is that we should always verify the reboot, according to this: https://github.com/mendersoftware/mender/commit/7fb71aab28edf3eed7799ddaca1d01ecd2cf10be

vpodzime: I don't care about the C++ client too much, TBH. This is a different world. In the MCU client, a reboot can mean merely restarting a thread or whatever else. So let's focus on the logic here. And I still see a logical error in the flow described above.

danielskinstad: Ah yes, verify reboot goes to verify reboot - I didn't catch that.  But does it not make sense that we always verify the reboot? Rollback reboot goes to verify rollback reboot - which verifies if it has been successful or not.
If it's not, it retries the rollback reboot - and this will continue until the verification is successful, or 
(when it's implemented) an infinite loop is detected.

vpodzime: > Ah yes, verify reboot goes to verify reboot - I didn't catch that.

Yes, that's a short loop. But also a failure in reboot goes to verify reboot -- why verify something that we already know failed?

danielskinstad: I guess my only motivation is to align it with the c++ client - i.e. follow the flow as closely as possible. What do you think @lluiscampos

lluiscampos: > I don't care about the C++ client too much

As much as possible, we need to align the behavior of the two clients. This is specially true for the state machine and the Update Modules API. We want is that a user familiar with Mender as a product, that has written maybe an Update Module at some point, can get started with Mender MCU faster and know what to expect.

Keep in mind also that our long term strategy is to merge the two state machines.

So yes, my vote on this one is to align with C++ client: ignore errors from the "do reboot" callback, and call "check reboot" callback.

lluiscampos (approved):

core/src/mender-client.c
vpodzime: Just to make it clear -- the current implementation will loop between a failing rollback reboot and a failing rollback reboot verification. With no sleep, nothing. When we add some counter/loop detection, it will just quickly exhaust the counter by calling the same callback(s) repeatedly.

lluiscampos: > but I'm not going to change the C++ client

Me neither.

> the current implementation will loop between a failing rollback reboot and a failing rollback reboot verification

That was an acknowledge typo, fixed now AFAICS.

(this also confused me in the first review, when I then suggested adding an sleep)

vpodzime: No, it's still there:
image

A failure in rollback reboot goes to verify rollback reboot which, because the reboot failed, is doomed to fail and goes back to rollback reboot. Or what am I missing here?

danielskinstad: You are right, and you are not missing anything. Rollback reboot will go to verify rollback reboot regardless. As I said, I just did this because it's how it is done the in the c++ client. It removed the failure state [here](https://github.com/mendersoftware/mender/commit/7fb71aab28edf3eed7799ddaca1d01ecd2cf10be). Seems like it's done this way to align it with the go client. I can agree that the behavior seems strange, but we want to align the state machines, right?

lluiscampos: >  is doomed to fail and goes back to rollback reboot. Or what am I missing here?

reboot -> verify -> reboot ==> That is correct. Eventually it will break out if we cannot reboot.

vpodzime: > Eventually it will break out if we cannot reboot.

That "eventually" will be after a sequence of calls of the two callbacks in a fast sequence with nothing in between them. Except for wearing the flash by writing the new state when switching between them.

vpodzime: > I can agree that the behavior seems strange, but we want to align the state machines, right?

Well, yes. But in the C++ client, the reboot, IIRC, is a code in the client itself. And it is a full OS reboot which is very unlikely to fail. Here we don't have a clue what the reboot callback does and how likely it is to fail.

vpodzime: Go ahead and merge this if you feel like it implements the correct behavior. There's no point in arguing whether it is or isn't correct. I just wanted to make sure you guys realize what the current implementation will do in case of a failing reboot (callback).

lluiscampos: I don't get it, @vpodzime. What is a "failed reboot" to you? A "no reboot"?

I think the whole reasoning is that the call might fail but the device is going to reboot anyway, so in my scenario there is no "fast sequence" between these two states.

vpodzime: Rollback reboot is a callback, right? And it might fail and perform no reboot. If that happens, we call the verify rollback reboot callback which should fail because there was no (rollback) reboot. And then we call the rollback reboot callback again while nothing has most likely changed since the last attempt unless the scheduler took out the CPU from us and let something else run. So it fails again. And then we call the verify rollback reboot callback which is should again fail (still no reboot). And this repeats basically in a spinlock until some counter or something is exhausted. In the the mean time switching between the states there and back wears flash.

lluiscampos: If it performed no reboot, the flow that you describe is correct.

The logic is based on the principle that reboot might return failure but still actually reboot, where then such a fast loop won't occur. Granted, this is based on Linux `reboot` command behavior and not applicable on a callback design (on the same binary!), were even an async reboot trigger of some sort should correctly return OK/FAIL.

Let's also look at this from higher level,

We are considering the rollback reboot case, no the regular reboot case where there shall not be retries that will end up in this infinite loop. The rollback reboot state is in the failure path, so it cannot be considered the "normal flow" and it will happen rarely. If a device is not able to rollback reboot, it is likely busted already anyway so wearing flash with ~10 more writes won't really make a difference I believe.

All in all I still think that keeping the two state machines aligned is the requirement that should prevail in this discussion.

vpodzime: Cool. As I wrote above, I'm not going to push you on that. I just wanted to make sure we acknowledged the "spin lock" behavior.

PR #93, opened 2024-10-11
chore: Rework how artifact download is handled

vpodzime:
In order to avoid passing many callbacks around as arguments and calling the update module's callback 9 levels of nesting deep, we just call the particular functions directly wherever possible. This requires a separate code path tailored to artifact download.

Ticket: MEN-7499
Changelog: none

### DIFF
--- a/core/src/mender-artifact.c
+++ b/core/src/mender-artifact.c
@@ -20,6 +20,7 @@
 #include <errno.h>
 
 #include "mender-artifact.h"
+#include "mender-client.h"
 #include "mender-log.h"
 
 /**
@@ -120,7 +121,7 @@ static mender_err_t mender_artifact_read_meta_data(mender_artifact_ctx_t *ctx);
  * @param callback Callback function to be invoked to perform the treatment of the data from the artifact
  * @return MENDER_DONE if the data have been parsed and payloads retrieved, MENDER_OK if there is not enough data to parse, error code if an error occurred
  */
-static mender_err_t mender_artifact_read_data(mender_artifact_ctx_t *ctx, mender_err_t (*callback)(char *, cJSON *, char *, size_t, void *, size_t, size_t));
+static mender_err_t mender_artifact_read_data(mender_artifact_ctx_t *ctx);
 
 /**
  * @brief Drop content of the current file of the artifact
@@ -194,13 +195,9 @@ is_compressed(const char *filename) {
 }
 
 mender_err_t
-mender_artifact_process_data(mender_artifact_ctx_t *ctx,
-                             void                  *input_data,
-                             size_t                 input_length,
-                             mender_err_t (*callback)(char *, cJSON *, char *, size_t, void *, size_t, size_t)) {
+mender_artifact_process_data(mender_artifact_ctx_t *ctx, void *input_data, size_t input_length) {
 
     assert(NULL != ctx);
-    assert(NULL != callback);
     mender_err_t ret = MENDER_OK;
     void        *tmp;
 
@@ -264,7 +261,7 @@ mender_artifact_process_data(mender_artifact_ctx_t *ctx,
             } else if (true == mender_utils_strbeginwith(ctx->file.name, "data")) {
 
                 /* Read data */
-                ret = mender_artifact_read_data(ctx, callback);
+                ret = mender_artifact_read_data(ctx);
 
             } else if (false == mender_utils_strendwith(ctx->file.name, ".tar")) {
 
@@ -849,10 +846,9 @@ mender_artifact_read_meta_data(mender_artifact_ctx_t *ctx) {
 }
 
 static mender_err_t
-mender_artifact_read_data(mender_artifact_ctx_t *ctx, mender_err_t (*callback)(char *, cJSON *, char *, size_t, void *, size_t, size_t)) {
+mender_artifact_read_data(mender_artifact_ctx_t *ctx) {
 
     assert(NULL != ctx);
-    assert(NULL != callback);
     mender_err_t ret;
 
     /* Retrieve payload index. We expect "data/%u.tar" where %u is the index.
@@ -884,7 +880,8 @@ mender_artifact_read_data(mender_artifact_ctx_t *ctx, mender_err_t (*callback)(c
     if (strlen("data/xxxx.tar") == strlen(ctx->file.name)) {
 
         /* Beginning of the data file */
-        if (MENDER_OK != (ret = callback(ctx->payloads.values[index].type, ctx->payloads.values[index].meta_data, NULL, 0, NULL, 0, 0))) {
+        if (MENDER_OK
+            != (ret = mender_client_download_artifact_callback(ctx->payloads.values[index].type, ctx->payloads.values[index].meta_data, NULL, 0, NULL, 0, 0))) {
             mender_log_error("An error occurred");
             return ret;
         }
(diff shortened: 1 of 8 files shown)


vpodzime:
This is what I was able to put together for https://northerntech.atlassian.net/browse/MEN-7499. However, it is here only for preview/discussion/suggestions purposes because it's quite wrong in at least two things:
- code duplication
- twisted dependencies between code units, e.g. `platform/net/zephyr/src/mender-http.c` having `#include "mender-api.h"` or `core/src/mender-artifact.c` having `#include "mender-client.h"`

And because I believe the above two issues are bad, especially the second one, I didn't bother adding the `posix` platform implementation of the `mender_http_artifact_download()` function, even though it would be easy because on POSIX it could just use the same callback passing as it does on the `main` branch.

Last but not least, this doesn't seem to fix the mysterious hangs and crashes. (although I keep tracking those down the :rabbit: :hole: !)

Feedback welcome!

vpodzime:
I hope to push this into a reasonable state during today and tomorrow (working from :car:)

larsewi (approved):
Great work 🚀  I know you mostly been moving code around, so feel free to dismiss my suggestions.

core/src/mender-artifact-download.c
```diff
@@ -0,0 +1,120 @@
+/**
+ * @file      mender-artifact-download.c
+ * @brief     Mender artifact download implementation
+ *
+ * Copyright Northern.tech AS
```
lluiscampos, line 5: Same question as for the other file.

core/src/mender-artifact.c
```diff
@@ -1119,15 +1187,16 @@ artifact_read_data(mender_artifact_ctx_t *ctx, mender_err_t (*callback)(char *,
...
-        if (MENDER_OK
-            != (ret = callback(ctx->payloads.values[index].type,
-                               ctx->payloads.values[index].meta_data,
-                               strstr(ctx->file.name, ".tar") + strlen(".tar") + 1,
-                               ctx->file.size,
-                               ctx->input.data,
-                               ctx->file.index,
-                               length))) {
+        /* Invoke the download artifact callback */
+        ret = process_artifact_data_callback(ctx->payloads.values[index].type,
```
lluiscampos, line 1191: From this line... such a simple change :smile:

platform/net/zephyr/src/mender-http.c
```diff
@@ -58,6 +60,14 @@ static mender_http_config_t mender_http_config;
  */
 static void mender_http_response_cb(struct http_response *response, enum http_final_call final_call, void *user_data);
 
+/**
+ * @brief HTTP artifact response callback, invoked to handle data received
+ * @param response HTTP response structure
+ * @param final_call Indicate final call
+ * @param user_data User data, used to retrieve request context data
+ */
+static void artifact_response_cb(struct http_response *response, enum http_final_call final_call, void *user_data);
```
lluiscampos, line 69: For consistency with the above cb: `mender_artifact_response_cb`

platform/net/zephyr/src/mender-http.c
```diff
@@ -257,6 +371,26 @@ mender_http_response_cb(struct http_response *response, enum http_final_call fin
     }
 }
 
+static void
+artifact_response_cb(struct http_response *response, MENDER_ARG_UNUSED enum http_final_call final_call, void *user_data) {
```
lluiscampos, line 375: Pre-existing,

Please add `MENDER_ARG_UNUSED enum...` to `mender_http_response_cb` also.

platform/net/zephyr/src/mender-http.c
```diff
@@ -257,6 +371,26 @@ mender_http_response_cb(struct http_response *response, enum http_final_call fin
...
+
+    assert(NULL != response);
+    assert(NULL != user_data);
+
+    /* Retrieve request context */
+    mender_artifact_download_data_t *dl_data = user_data;
```
lluiscampos, line 381: We should cast the void pointer to `(mender_artifact_download_data_t *) user_data`

lluiscampos (approved):
Great work :rocket:  Really nice commit message, thank you.

platform/net/zephyr/src/mender-http.c
vpodzime: That's done by the assignment.

lluiscampos: TIL! Thanks

core/src/mender-artifact-download.c
```diff
@@ -0,0 +1,120 @@
...
+#include "mender-http.h"
+#include "mender-log.h"
+
+static mender_err_t mender_download_artifact_callback(mender_http_client_event_t       event,
+                                                      void                            *data,
+                                                      size_t                           data_length,
+                                                      mender_artifact_download_data_t *dl_data);
+
+mender_err_t
+mender_download_artifact(char *uri, mender_deployment_data_t *deployment_data, mender_update_module_t **update_module) {
```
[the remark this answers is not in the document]
vpodzime: Yup, exactly my thinking. But then I saw I'd have to change the functions this is passed into... Will address this in a fw-up.

include/mender-http-client-event.h
```diff
@@ -0,0 +1,42 @@
+/**
+ * @file      mender-http-client-event.h
+ * @brief     Mender HTTP client event type
+ *
+ * Copyright joelguittet and mender-mcu-client contributors
```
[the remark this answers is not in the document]
vpodzime: Yes, as long as they contain Joel's code.

core/src/mender-update-module.c
```diff
@@ -0,0 +1,77 @@
+/**
+ * @file      mender-update-module.c
+ * @brief     Mender update Module implementation
+ *
+ * Copyright Northern.tech AS
```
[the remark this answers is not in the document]
vpodzime: Yup! Good catch!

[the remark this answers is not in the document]
vpodzime: Actually, not really. I wrote this code for update modules, right?

core/src/mender-artifact-download.c
vpodzime: This one has Joel's code, that's true. Adding.

core/src/mender-artifact.c
vpodzime: ?

platform/net/zephyr/src/mender-http.c
vpodzime: Actually, I think it should be the other way around -- no `mender_` prefix for `static` functions.

vpodzime:
@larsewi @lluiscampos I addressed the issues introduced in this PR and I'm merging it as-is now. Will follow up with a PR addressing the unresolved pre-existing issues you pointed out.

include/mender-client.h
```diff
@@ -137,6 +129,21 @@ mender_err_t mender_client_ensure_authenticated(void);
...
+ * @param type Type from header-info payloads
+ * @param meta_data Meta-data from header tarball
+ * @param filename Artifact filename
+ * @param size Artifact file size
+ * @param data Artifact data
+ * @param index Artifact data index
+ * @param length Artifact data length
+ * @return MENDER_OK if the function succeeds, error code if an error occurred
+ */
+mender_err_t mender_client_download_artifact_callback(char *type, cJSON *meta_data, char *filename, size_t size, void *data, size_t index, size_t length);
```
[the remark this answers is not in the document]
vpodzime: This should no longer exist, actually.

core/src/mender-update-module.c
```diff
@@ -0,0 +1,77 @@
...
+mender_update_module_t *
+mender_update_module_get(const char *artifact_type) {
+    mender_update_module_t *ret = NULL;
+
+    /* Treatment depending of the type */
+    if (NULL != update_modules_list) {
+        for (size_t update_module_index = 0; (NULL == ret) && (update_module_index < update_modules_count); update_module_index++) {
+            /* Check artifact type */
+            if (StringEqual(artifact_type, update_modules_list[update_module_index]->artifact_type)) {
+                ret = update_modules_list[update_module_index];
```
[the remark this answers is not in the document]
vpodzime: > Shouldn't we `break` here?

The loop condition checks `ret` and stops once it's set.

> Or amend the registering function?

Yes, I think we should not allow multiple UMs for the same artifact type.

core/src/mender-artifact.c
lluiscampos: It was meant a compliment. If you only look at this line your PR seems "so simple", yet the work is everywhere else to make this possible :rocket:

PR #94, opened 2024-10-15
Two memory management improvements

vpodzime:


core/src/mender-api.c
```diff
@@ -717,7 +717,11 @@ mender_api_http_artifact_callback(mender_http_client_event_t event, void *data,
     switch (event) {
         case MENDER_HTTP_EVENT_CONNECTED:
             /* Create new artifact context */
-            if (NULL == (mender_artifact_ctx = mender_artifact_create_ctx())) {
+            /* Having the context's internal buffer of size (2 * MENDER_ARTIFACT_STREAM_BLOCK_SIZE +
+               mender_http_recv_buf_length) means we should always have enough space for unprocessed
```
vpodzime, line 721: Actually, this is not always true it seems. (clever me adding that log message when growing the buffer! :grin: :innocent: )

larsewi (approved):
:rocket:

vpodzime:
I created https://northerntech.atlassian.net/browse/MEN-7618 for the unnecessary spike in memory usage for `header.tar`.

core/src/mender-artifact.c
```diff
@@ -300,14 +303,37 @@ mender_artifact_process_data(mender_artifact_ctx_t *ctx,
...
 
     /* Copy data to the end of the internal buffer */
     if ((NULL != input_data) && (0 != input_length)) {
-        if (NULL == (tmp = realloc(ctx->input.data, ctx->input.length + input_length))) {
-            /* Unable to allocate memory */
-            return MENDER_FAIL;
+        if ((ctx->input.length + input_length) > ctx->input.size) {
+            new_size = ctx->input.length + input_length;
+            /* data/ files are processed per block for which the original size of the buffer should
+               be enough, but metadata is processed as whole files so there we except we will need
```
danielskinstad, line 314:
```suggestion
               be enough, but metadata is processed as whole files so there we expect we will need
```

danielskinstad (approved):
Looks good as far as I can tell

lluiscampos (commented):
Looks good :+1:

PR #95, opened 2024-10-15
fix: Introduce mender_utils_hexdump_to_bytes() and use it in checksum verification

vpodzime:
We have seen sscanf() causing hard failures on real HW, at least on the ESP32-based ones. Let's avoid it in favor of our simple and fast code.

Ticket: MEN-7562
Changelog: none


larsewi (approved):
Thanks for cleaning up my mess :rocket:

lluiscampos:
It looks good. Thank you :+1:

PR #96, opened 2024-10-17
feat: infinite state loop detection

danielskinstad:
Add transition for state loop and exit if MAX_STATE_DATA_STORE_COUNT is reached.

Ticket: MEN-7516
Changelog: Implement infinite state loop detection

### DIFF
--- a/core/src/mender-client.c
+++ b/core/src/mender-client.c
@@ -899,18 +899,20 @@ set_and_store_state(const mender_update_state_t state) {
      * Set the state in `mender_client_deployment_data` and write it to the nvs
      */
 
+    mender_err_t ret = MENDER_OK;
+
     /* Set state in deployment data */
-    if (MENDER_OK != mender_deployment_data_set_state(mender_client_deployment_data, state)) {
+    if (MENDER_OK != (ret = mender_deployment_data_set_state(mender_client_deployment_data, state))) {
         mender_log_error("Failed to set deployment data state");
-        return MENDER_FAIL;
+        return ret;
     }
 
     /* Store deployment data */
-    if (MENDER_OK != mender_set_deployment_data(mender_client_deployment_data)) {
+    if (MENDER_OK != (ret = mender_set_deployment_data(mender_client_deployment_data))) {
         mender_log_error("Failed to store deployment data");
-        return MENDER_FAIL;
+        return ret;
     }
-    return MENDER_OK;
+    return ret;
 }
 
 static mender_err_t
@@ -964,13 +966,16 @@ mender_client_update_work_function(void) {
         update_state = update_state_transitions[update_state].success;         \
         assert(NULL != mender_update_module);                                  \
         mender_log_debug("Entering state %s", update_state_str[update_state]); \
-        set_and_store_state(update_state);                                     \
-        ret = MENDER_OK;                                                       \
+        if (MENDER_LOOP_DETECTED == set_and_store_state(update_state)) {       \
+            update_state = MENDER_UPDATE_STATE_FAILURE;                        \
+        }                                                                      \
     } else {                                                                   \
         update_state = update_state_transitions[update_state].failure;         \
         mender_log_debug("Entering state %s", update_state_str[update_state]); \
         if (NULL != mender_update_module) {                                    \
-            set_and_store_state(update_state);                                 \
+            if (MENDER_LOOP_DETECTED == set_and_store_state(update_state)) {   \
+                update_state = MENDER_UPDATE_STATE_FAILURE;                    \
+            }                                                                  \
         }                                                                      \
         ret = MENDER_OK;                                                       \
         continue;                                                              \
@@ -1154,10 +1159,8 @@ mender_client_update_work_function(void) {
                     /* If the rollback verify reboot fails,
                      * we will retry the rollback reboot.
                      *
-                     * TODO: fix this comment / document infinite
-                     * loop detection once MEN-7516 is implemented.
-                     * As it is now, the rollback will be retried
-                     * infinitely, or until it succeeds
+                     * The `rollback-reboot -> rollback-verify-reboot -> rollback-reboot -> ...`
+                     * loop is broken when a state loop is detected
                      */
                     mender_log_error("Rollback verify reboot failed. Retry rollback reboot");
                 }
--- a/core/src/mender-deployment-data.c
+++ b/core/src/mender-deployment-data.c
@@ -33,8 +33,12 @@
  * @brief Deployment data version number.
  * @note cJSON stores numbers as double, so we might as well define this
  *       constant as a double to avoid type casting.
+ *       This number is based on the maximum state transition count
+ *       from the Mender C++ client. It's a number that won't loop
+ *       too many times and still give some wiggle room.
+ *       If this number is exceeded, the update will be aborted.
  */
-#define MAX_STATE_DATA_STORE_COUNT 50.0 /* TODO: What should this constant be? */
+#define MAX_STATE_DATA_STORE_COUNT 28.0
 
 /**
  * @brief Validate deployment data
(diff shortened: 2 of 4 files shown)


include/mender-utils.h
```diff
@@ -91,6 +91,7 @@ typedef enum {
     MENDER_FAIL            = -1, /**< Failure */
     MENDER_NOT_FOUND       = -2, /**< Not found */
     MENDER_NOT_IMPLEMENTED = -3, /**< Not implemented */
+    MENDER_STATE_LOOP      = -4, /**< State loop */
```
lluiscampos, line 94: It is an error, I would name it with something more explicit. Examples:
* `MENDER_STATE_LOOP_DETECTED` (this is similar to the C++ error name)
* `MENDER_MAX_STATE_COUNT_EXCEEDED`
* ...

lluiscampos: Also, I don't really like exposing this error in `mender-utils.h`. But this is kind of the issue with no proper namespaces as the project is structured :cry:

platform/tls/generic/mbedtls/src/mender-tls.c
```diff
@@ -192,6 +192,9 @@ mender_tls_init_authentication_keys(mender_err_t (*get_user_provided_keys)(char
         case MENDER_NOT_IMPLEMENTED:
             assert(false && "Unexpected return value");
             /* fallthrough */
+        case MENDER_STATE_LOOP:
+            assert(false && "Unexpected return value");
+            /* fallthrough */
```
lluiscampos, line 197: Why do we need to care about this case here? I don't get it.

core/src/mender-deployment-data.c
```diff
@@ -108,7 +108,7 @@ mender_set_deployment_data(mender_deployment_data_t *deployment_data) {
     assert(NULL != item); /* Validation above should have catched this already */
     if (MAX_STATE_DATA_STORE_COUNT <= cJSON_GetNumberValue(item)) {
```
lluiscampos, line 109: In this task, we also need to get a sensible value for `MAX_STATE_DATA_STORE_COUNT`. We should document (in the code) why we chose such a number.

In the C++ project we have this:
```
// The maximum times we are allowed to move through update states. If this is exceeded then the
// update will be forcefully aborted. This can happen if we are in a reboot loop, for example.
const int kMaxStateDataStoreCount = 28;
```

I'll ask Kristian about the guideline for this number.

core/src/mender-client.c
```diff
@@ -83,23 +83,26 @@ mender_client_state_t mender_client_state = MENDER_CLIENT_STATE_INITIALIZATION;
...
+    /* MENDER_UPDATE_STATE_VERIFY_REBOOT          */ { MENDER_UPDATE_STATE_COMMIT, MENDER_UPDATE_STATE_ROLLBACK, MENDER_UPDATE_STATE_FAILURE },
+    /* MENDER_UPDATE_STATE_COMMIT                 */ { MENDER_UPDATE_STATE_CLEANUP, MENDER_UPDATE_STATE_ROLLBACK, MENDER_UPDATE_STATE_FAILURE },
+    /* MENDER_UPDATE_STATE_CLEANUP                */ { MENDER_UPDATE_STATE_END, MENDER_UPDATE_STATE_END, MENDER_UPDATE_STATE_END },
+    /* MENDER_UPDATE_STATE_ROLLBACK               */ { MENDER_UPDATE_STATE_ROLLBACK_REBOOT, MENDER_UPDATE_STATE_FAILURE, MENDER_UPDATE_STATE_FAILURE },
+    /* MENDER_UPDATE_STATE_ROLLBACK_REBOOT        */ { MENDER_UPDATE_STATE_ROLLBACK_VERIFY_REBOOT, MENDER_UPDATE_STATE_ROLLBACK_VERIFY_REBOOT, MENDER_UPDATE_STATE_FAILURE },
+    /* MENDER_UPDATE_STATE_ROLLBACK_VERIFY_REBOOT */ { MENDER_UPDATE_STATE_FAILURE, MENDER_UPDATE_STATE_ROLLBACK_REBOOT, MENDER_UPDATE_STATE_FAILURE },
+    /* MENDER_UPDATE_STATE_FAILURE                */ { MENDER_UPDATE_STATE_CLEANUP, MENDER_UPDATE_STATE_CLEANUP, MENDER_UPDATE_STATE_CLEANUP },
```
lluiscampos, line 103: I'll let @vpodzime comment on this part - maybe it is not worth to have a third parameter in the transition struct if it's only two states that have something different than `FAILURE` (iow maybe is better to have to if/else if for these)

lluiscampos (approved):
I am surprised on how easy this turned out :rocket: I _think_  that it works correctly (and I thanks for the demo on your desk!)

platform/tls/generic/mbedtls/src/mender-tls.c
danielskinstad: The smoke build threw errors that the switch had unhandled enumeration values.

larsewi (approved):
Great :rocket:

core/src/mender-deployment-data.c
danielskinstad: From the [commit](https://github.com/mendersoftware/mender/commit/12e6c62302e6dcca51d3efb382e276515552b718) that added it: 
`There is nothing accurate about this number. It's just a number which
won't loop too many times, but still gives some wiggle room to add a
few more states without exceeding the number. It was determined simply
by matching the number of loops in one of the tests which tries to
exceed it.`

include/mender-utils.h
```diff
@@ -86,11 +86,12 @@ extern "C" {
...
-    MENDER_OK              = 0,  /**< OK */
-    MENDER_FAIL            = -1, /**< Failure */
-    MENDER_NOT_FOUND       = -2, /**< Not found */
-    MENDER_NOT_IMPLEMENTED = -3, /**< Not implemented */
+    MENDER_DONE                = 1,  /**< Done */
+    MENDER_OK                  = 0,  /**< OK */
+    MENDER_FAIL                = -1, /**< Failure */
+    MENDER_NOT_FOUND           = -2, /**< Not found */
+    MENDER_NOT_IMPLEMENTED     = -3, /**< Not implemented */
+    MENDER_STATE_LOOP_DETECTED = -4, /**< State loop detected */
```
vpodzime, line 94: I'd call this `MENDER_LOOP_DETECTED`. These values should be generic.

core/src/mender-client.c
vpodzime: Yes. And there is no way the two special states will loop, right?

vpodzime (commented):
Looks good to me otherwise.

core/src/mender-client.c
danielskinstad: They won't loop - all states except `failure` and `cleanup` goes to `failure`; `failure` goes to `cleanup`, and `cleanup` always goes to `end`. But you don't think it should be a part of the transition struct? Isn't it cleaner/clearer if we have the possible transitions in one place?

vpodzime: I'm thinking a loop should always be quit by going to the `failure` state with the counter reset. That way we hopefully report an error to the server, do a cleanup and return back to normal operation.

danielskinstad: Right - so just `if loop-detected: update_state = failure` then I would agree that it's not necessary to have it in the struct - if that's what you mean

vpodzime: > there is no point in going to failure if we're in the cleanup state

Right, but there can never be a loop when we entered the `cleanup` state, right? And the same applies to `failure`. That's what in meant [above](https://github.com/mendersoftware/mender-mcu/pull/96#discussion_r1808335109)

danielskinstad: Maybe I am misunderstanding you, but I do believe it's possible that there somehow is a back-and-forth between some states which can result in cleanup being entered right at the state-loop threshold, resulting in a loop being detected while we're entering cleanup. Same with failure - since both states call NEXT_STATE, which attempts to store the deployment data and increases the state loop counter. So we would probably need to store which state we're in before calling transition.success or transition.failure, then have logic to not enter failure in failure and cleanup. It isn't hard to implement, but I feel like that creates more clutter in `NEXT_STATE` rather than simply having a third column in the transition struct :sweat_smile:  - or maybe I am just overlooking a much simpler solution

vpodzime: > I do believe it's possible that there somehow is a back-and-forth between some states which can result in cleanup being entered right at the state-loop threshold
Then we have a bad threshold because it's not a loop, right? But if we just go to `failure` every time we detect a loop and reset the counter, the worst that can happen is that we:
1. detect a loop while entering `cleanup`
2. go to `failure` and reset the counter
3. go to `cleanup` with counter at 1

(or the same for detecting loop when entering `failure`).

Is that not going to work in some case? `failure` an `cleanup` cannot create a loop on their own, right?

danielskinstad: That will indeed work - didn't even think about just resetting the counter, thanks :)

vpodzime (approved):
Nice and easy! Thanks! :clap:

PR #97, opened 2024-10-17
MEN-7526: Introduce Unit Test framework

lluiscampos:
Refactor of the build structure and introduction of the Unit Test framework (together with one test thanks to @danielskinstad)

Still few things on my plate :spaghetti: 

- [x] Remove proposal document - this was just meant for the demo
- [ ] Clear TODOs (either do them or create follow-up tasks)
- [ ] Align cJSON version with the one used for Zephyr builds
- [x] Run tests in CI
- [x] Add coverage
- [x] (maybe) Run tests with different sanitizers
- [x] Publish coverage in coveralls


lluiscampos:
@vpodzime @danielskinstad There are several things that I still want to address but they will be coming in separate commits - hence you are welcome to review what I did so far :+1:

tests/unit/core/mender-utils_test.cpp
```diff
@@ -0,0 +1,75 @@
...
+#include <gtest/gtest.h>
+
+extern "C" {
+#include "mender-utils.h"
+#include "mender-client.h"
+
+#include <stdarg.h>
+#include <stddef.h>
+#include <setjmp.h>
+}
```
vpodzime, line 29: This seems wrong. Our headers should not require this and the standard ones should be used in the C++ way, right?

vpodzime (commented):
Looks good to me otherwise. Thanks for the big chunk of work on this! One thing I have thought about over the weekend -- isn't the C++ compiler potentially going to create different binaries from our code than a C compiler would? Could we compile our code using a C compiler and then only the tests using a C++ compiler?

tests/unit/CMakeLists.txt
```diff
@@ -0,0 +1,127 @@
...
+include(FetchContent)
+FetchContent_Declare(
+  cjson
+  GIT_REPOSITORY https://github.com/DaveGamble/cJSON
+  GIT_TAG        v1.7.15
+  DOWNLOAD_EXTRACT_TIMESTAMP TRUE
+)
+
+set(ENABLE_CJSON_TEST OFF CACHE INTERNAL "Turn off tests")
+FetchContent_MakeAvailable(cjson)
```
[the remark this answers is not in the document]
lluiscampos: I haven't decided yet, thanks for bringing it up.

We should either expect system provided libraries or fetch them, but not both. The old code already had the flaw that in the top `CMakeLists.txt` would use the system ones, while the `tests/CMakeLists.txt` one (now in `tests/smoke/...`) would fetch them from the "mocks" (I am talking plural because the same is true for MbedTLS). I think then that depending on how you run your CMake command you will get one or the other. Another related flaw was the (ab)use of `include` function on `CMakeLists.txt`.

These issues are still present for the smoke tests build.

I think we should flavor system ones in this repository, and unify the checks for it in the `target/posix/CMakeLists.txt`.

tests/unit/core/mender-utils_test.cpp
lluiscampos: Thanks, I will look into it :thinking:

tests/unit/CMakeLists.txt
```diff
@@ -0,0 +1,125 @@
...
+  )
+endif()
+
+add_custom_target(coverage
+  COMMAND lcov --capture --quiet --directory .
+               --output-file coverage.lcov
+               --exclude '/usr/*'
+               --exclude '*/googletest/*'
+               --exclude '*_test.*'
+#               --exclude '*/googlemock/*'
```
danielskinstad, line 90: Why is this commented out?

danielskinstad (approved):
Nice work!

tests/unit/CMakeLists.txt
lluiscampos: We will need when adding tests that make user of it, for example with `EXPECT_THAT` type of macros. But I think I can leave it un-commented, actually - it shouldn't matter that the pattern does not match.

tests/unit/core/mender-utils_test.cpp
lluiscampos: They were not needed - I removed them.

tests/unit/CMakeLists.txt
lluiscampos: Actually not :laughing: 

```
geninfo: ERROR: 'exclude' pattern '*/googlemock/*' is unused.
```

I will leave it around if you don't mind. It will come in handy when writing tests, I am pretty sure.

tests/unit/CMakeLists.txt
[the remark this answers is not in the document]
lluiscampos: I'll follow-up in the next PR

lluiscampos:
To not make this PR too long (and to verify the base coverage for `main` to compare against) I'll merge this PR as-is. I'll follow up with another one with:

>     * [ ]  Clear TODOs (either do them or create follow-up tasks)
>     * [ ]  Align cJSON version with the one used for Zephyr builds

PR #100, opened 2024-10-24
MEN-7517: Ensure atomicity/consistency of saved state data

danielskinstad:
This is almost the same as just enabling the CRC option, so my only motivation for doing this is that it might be possible that you for some reason don't want a CRC on all the data in the NVS? If that's not a concern, the alternative to this is to simply enable `CONFIG_NVS_DATA_CRC` , which will then write and verify a CRC for all data elements in the nvs.

### DIFF
--- a/platform/storage/zephyr/nvs/src/mender-storage.c
+++ b/platform/storage/zephyr/nvs/src/mender-storage.c
@@ -21,6 +21,7 @@
 #include <errno.h>
 #include <zephyr/drivers/flash.h>
 #include <zephyr/fs/nvs.h>
+#include <zephyr/sys/crc.h>
 #include <zephyr/storage/flash_map.h>
 #include "mender-log.h"
 #include "mender-storage.h"
@@ -82,6 +83,50 @@ checked_nvs_write(struct nvs_fs *fs, uint16_t id, const void *data, size_t len) 
      *    When a rewrite of the same data already stored is attempted, nothing is written to flash, thus 0 is returned.
      */
     return (len == ret) || (0 == ret);
+}
+
+static mender_err_t
+crc_add(char **data, size_t *data_len) {
+
+    assert(NULL != data);
+    assert(NULL != data_len);
+
+    uint32_t crc = crc32_ieee(*data, *data_len);
+    char    *tmp = realloc(*data, *data_len + sizeof(crc));
+    if (NULL == tmp) {
+        mender_log_error("Unable to allocate memory for deployment data");
+        return MENDER_FAIL;
+    }
+    memcpy(tmp + *data_len, &crc, sizeof(crc));
+    *data_len += sizeof(crc);
+    *data = tmp;
+
+    return MENDER_OK;
+}
+
+static mender_err_t
+crc_check(const unsigned char *data, const size_t data_len) {
+
+    assert(NULL != data);
+
+    if (data_len > sizeof(uint32_t)) {
+
+        /* Extract the CRC  */
+        uint32_t crc = *(uint32_t *)(data + data_len - sizeof(crc));
+
+        /* Compute CRC of the loaded data */
+        uint32_t computed_crc = crc32_ieee(data, data_len - sizeof(crc));
+
+        if (computed_crc != crc) {
+            mender_log_error("CRC mismatch in deployment data");
+            return MENDER_FAIL;
+        }
+    } else {
+        mender_log_error("Invalid deployment data size (too small for CRC)");
+        return MENDER_FAIL;
+    }
+
+    return MENDER_OK;
 }
 
 mender_err_t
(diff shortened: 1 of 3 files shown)


platform/storage/zephyr/nvs/src/mender-storage.c
```diff
@@ -207,6 +261,12 @@ mender_storage_get_deployment_data(char **deployment_data) {
         return ret;
     }
 
+#ifdef CONFIG_MENDER_STORAGE_DEPLOYMENT_DATA_CRC
+    if (MENDER_OK != check_crc(*deployment_data, deployment_data_length)) {
+        return MENDER_FAIL;
+    }
+#endif /* CONFIG_MENDER_STORAGE_DEPLOYMENT_DATA_CRC */
```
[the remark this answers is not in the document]
danielskinstad: We can still parse it as JSON, the CRC is added after the null-terminator, right? I have tested this, and it seems to work fine

platform/storage/zephyr/nvs/src/mender-storage.c
```diff
@@ -207,6 +261,12 @@ mender_storage_get_deployment_data(char **deployment_data) {
         return ret;
     }
 
+#ifdef CONFIG_MENDER_STORAGE_DEPLOYMENT_DATA_CRC
+    if (MENDER_OK != check_crc(*deployment_data, deployment_data_length)) {
+        return MENDER_FAIL;
```
lluiscampos, line 266: What should we do if the crc is invalid?

As I read the code, what I see is that `mender_get_deployment_data` will return FAIL in `mender_client_initialization_work_function` and then go to REBOOT. It could be that we are in an state where reboot will bring us back to a working image (for example with MCU Boot, etc) but it could also be that it is not the case and we will keep rebooting forever.

I think that the safest is to erase the deployment data and let the deployment start again. Or? @vpodzime ?

What will happen to the state machine when we return `MENDER_FAIL` -> I suspect it will just go into the fail states path, but still won't be able to keep saving state and then the

lluiscampos: Also, related but clearly scope creep, this also applies if you are not able to read the private key from NVS at start up -> what should we do then?

(I am bringing this one in hoping to ease the discussion)

lluiscampos (approved):
The code looks good.

I have a high level question that we should think about :thinking:

platform/storage/zephyr/nvs/src/mender-storage.c
danielskinstad: That doesn't only happen if the CRC is invalid, though. It happens if the parsing fails or if the deployment data is invalid as well. We only call `mender_get_deployment_data` on a boot, right? So after a reboot on a deployment, if the deployment data for some reason is broken, isn't it perhaps safer to return back to the partition we started the deployment in?

platform/storage/zephyr/nvs/src/mender-storage.c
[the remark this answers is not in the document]
vpodzime: Cool. I was worried cJSON would complain about the garbage after the JSON data. But I guess it just doesn't care.

platform/storage/zephyr/nvs/src/mender-storage.c
vpodzime: This is partially related to MEN-7555. I think we should log an error, reboot back into the working image and if the deployment data is not valid in there either, we should delete it. The biggest issue I see is that without deployment data we don't know for which deployment to log this error and report the failure status. So the deployment will be left in-progress on the server. If it **is** valid in the previous image, we should report a failure and the error.

platform/storage/zephyr/nvs/src/mender-storage.c
```diff
@@ -181,8 +227,16 @@ mender_storage_set_deployment_data(char *deployment_data) {
...
+    size_t data_len = strlen(deployment_data) + 1;
+
+#ifdef CONFIG_MENDER_STORAGE_DEPLOYMENT_DATA_CRC
+    if (MENDER_OK != add_crc(&deployment_data, &data_len)) {
+        return MENDER_FAIL;
+    }
+#endif /* CONFIG_MENDER_STORAGE_DEPLOYMENT_DATA_CRC */
+
+    /* Write deployment data  */
+    if (nvs_write(&mender_storage_nvs_handle, MENDER_STORAGE_NVS_DEPLOYMENT_DATA, deployment_data, data_len) < 0) {
```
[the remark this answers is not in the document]
vpodzime: Well, it would make the code simpler in a way by not having to do `realloc()` and point `uint32_t` to somewhere after string data, etc. But YMMV. The only functional difference is that by writing the CRC32 first, you immediately invalidate the old deployment data (if any) and so after a sudden reboot it will not be considered valid. However, I don't see any case where this would make a practical difference in our use case, TBH.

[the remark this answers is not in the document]
danielskinstad: By appending it to the end of the data, we could reuse the functions if we want to apply it to other data, that way we can avoid the need for a separate key per CRC

platform/storage/zephyr/nvs/src/mender-storage.c
lluiscampos: One note: remember that not always will be a second image, some Update Modules won't be "firmware update" but changing something in the current image, or in an external memory, or whatever that does not require reboot.

But okey, we can address this with MEN-7555 :+1: Thanks for the ideas

platform/storage/zephyr/nvs/src/mender-storage.c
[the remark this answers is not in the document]
vpodzime: True.

platform/storage/zephyr/nvs/src/mender-storage.c
```diff
@@ -84,6 +85,50 @@ checked_nvs_write(struct nvs_fs *fs, uint16_t id, const void *data, size_t len)
...
+    }
+    memcpy(tmp + *data_len, &crc, sizeof(crc));
+    *data_len += sizeof(crc);
+    *data = tmp;
+
+    return MENDER_OK;
+}
+
+static mender_err_t
+crc_check(const char *data, const size_t data_len) {
```
vpodzime, line 108: It would be better (more correct) to use `unsigned char` here. Or `uint8_t`.

vpodzime (approved):
Just a nitpick. ACK.

PR #101, opened 2024-10-24
MEN-7526: Unit Test framework follow-ups

lluiscampos:
Clean-ups and follow-ups from #97


vpodzime (commented):
+ ~loner~ longer in the last commit message.

Looks good to me otherwise.

.gitlab-ci.yml
```diff
@@ -43,17 +45,23 @@ test:smoke-build:weak:
...
     - cmake --build build --parallel $(nproc --all)
     - ./build/mender-mcu-client.elf --help
 
 test:unit:
   stage: test
-  image: debian:12
+  image: debian:12-slim
   before_script:
-    - apt update && apt install -yyq g++ cmake git make lcov pkg-config curl
+    - apt update && apt install -yyq
```
danielskinstad, line 60: I don't think it matters much, but isn't `apt-get` preferred for scripts?

cmake/asan.cmake
```diff
@@ -0,0 +1,24 @@
+# @file      asan.cmake
+# @brief     Address Sanitazier options
```
danielskinstad, line 2:
```suggestion
# @brief     Address Sanitizer options
```

cmake/helper.cmake
```diff
@@ -0,0 +1,39 @@
+# @file      helper.cmake
+# @brief     Helper for sanitaziers
```
danielskinstad, line 2:
```suggestion
# @brief     Helper for sanitizers
```

cmake/threadsan.cmake
```diff
@@ -1,7 +1,7 @@
-# @file      CMakeLists.txt
-# @brief     Mocks CMakeLists file
+# @file      threadsan.cmake
+# @brief     Thread Sanitazier options
```
danielskinstad, line 2:
```suggestion
# @brief     Thread Sanitizer options
```

cmake/undefsan.cmake
```diff
@@ -0,0 +1,24 @@
+# @file      undefsan.cmake
+# @brief     Undefined Sanitazier options
```
danielskinstad, line 2:
```suggestion
# @brief     Undefined Sanitizer options
```

danielskinstad (approved):

tests/unit/CMakeLists.txt
```diff
@@ -36,9 +36,16 @@ include_directories(${CMAKE_CURRENT_BINARY_DIR}/_deps/cjson-src/)
...
-# Include mender-mcu-client library
+# Include mender-mcu-client library, forcing warnings and errors
+set(CONFIG_MENDER_ALL_WARNINGS_AS_ERRORS ON CACHE BOOL "All warnings as errors" FORCE)
 add_subdirectory("${CMAKE_CURRENT_LIST_DIR}/../../target/posix" build)
 
+# And add the same flags for the test targets too
+add_compile_options(-Wall -Wextra -Werror -Wpointer-arith)
+
+# Silence C/C++ compat warning for structs and unions
+add_compile_options(-Wno-extern-c-compat)
```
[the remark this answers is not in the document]
lluiscampos: Of course. Thank you.

target/posix/CMakeLists.txt
```diff
@@ -74,14 +74,29 @@ if (CONFIG_MENDER_PROVIDES_DEPENDS)
...
+
+# Depencency: Mbed TLS. Fetch the project from upstream
+# This is because the popular Linux distributions (Fedora, Ubuntu) are stuck in 2.x version
+# due to an breaking API change in 3.x. As Zephyr project uses 3.x, we want to avoid poluting
+# the code with many ifdefs for the two major versions
+include(FetchContent)
+FetchContent_Declare(
+  MbedTLS
+  GIT_REPOSITORY https://github.com/Mbed-TLS/mbedtls
+  GIT_TAG        v3.6.0
```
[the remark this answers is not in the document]
lluiscampos: Here I sneaked in my personal preference of aligning it to Zephyr project... :see_no_evil: 

To be honest, I don't have an strong opinion on this. Having a newer one in the same series should still be fully compatible. So, once you commented on it, I'll update it :grin:

tests/smoke/CMakeLists.txt
```diff
@@ -75,10 +72,10 @@ endif()
 
 # Link the mender-mcu-client library
 if(CONFIG_MENDER_PLATFORM_NET_TYPE MATCHES "generic/curl")
-    target_link_libraries(mender-mcu-client curl)
+    target_link_libraries(mender-mcu-client PRIVATE curl)
 endif()
 if(CONFIG_MENDER_PLATFORM_SCHEDULER_TYPE MATCHES "posix")
-    target_link_libraries(mender-mcu-client rt)
+    target_link_libraries(mender-mcu-client PRIVATE rt)
```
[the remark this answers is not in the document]
lluiscampos: To be honest this I don't know where is coming from - is it really the library or the smoke tests? If it is the library, should we move it into the library's `CMakeLists.txt` file?

This part (the smoke tests) I didn't want to rework more than necessary. It has an error (from CMake point of view) in that in uses the `mender-mcu-client` target directly, where what it should do is to have its own target and then link to the library (similar as it is done in the unit tests).

(the above is rather a rant than a reply to you - if you confirm that this is part of the library I'll modify it as public :ok_hand: )

target/posix/CMakeLists.txt
```diff
@@ -22,6 +22,7 @@ cmake_minimum_required(VERSION 3.16.3)
 project(mender-mcu-client LANGUAGES C)
 
 # Include helper to define MENDER_MCU_ variables
+include(../../cmake/CMake_posix_defaults.txt)
```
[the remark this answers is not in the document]
lluiscampos: Yes, thanks.

.gitlab-ci.yml
lluiscampos: It is. But we have this everywhere already. I believe folks in Ubuntu won't dare to change this now after many years...

cmake/undefsan.cmake
lluiscampos: Thanks for all the typos :see_no_evil:

tests/smoke/CMakeLists.txt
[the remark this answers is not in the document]
vpodzime: Oh! Wait, this is wrong in a different way then, right? To me it looks like this should be in `target/posix/CMakeLists.txt` as a `PUBLIC` dependency of the library when building for POSIX, right?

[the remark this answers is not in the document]
lluiscampos: Exactly!

I won't fix it in this PR, though. But we can revisit the smoke tests _target_ on a later point to correct it.

PR #102, opened 2024-10-25
chore: validate state in deployment data

danielskinstad:
Was overlooked when adding state to the deployment data.

Ticket:None


lluiscampos (approved):

vpodzime (approved):
ACK

PR #103, opened 2024-10-25
chore: remove "this is a bug" from debug log

danielskinstad:
It's not a bug to be in `MENDER_UPDATE_STATE_END`. We transition here from cleanup, so we will always reach this state.


lluiscampos (approved):
At the time I wrote this it really was a bug :lady_beetle: :laughing:

PR #104, opened 2024-10-31
MEN-7536: Scheduler rework -- part one, cleanup

vpodzime:
The first part of the scheduler rework. It doesn't bring any of the features we need, but it's a cleanup of the current state which should hopefully be a good starting point for starting to add the new features.


vpodzime:
I decided to submit this alone because it's already big enough and worth reviewing, I believe. I left the mutex code intact because it will be needed for the extra features to come:

> we need to add support for:
> *    Inventory (most likely yet another delayable work)
> *    Correct handling of poll/inventory intervals 
> *    Thread safety
> *    Need to test kernel work queue vs user-defined work queue
> *    Need to add support for interrupting tasks in the middle of execution

lluiscampos:
@vpodzime I hope that @pasinskim can review this one. If he does not get on it today, I'll review it on my next GitHub PRs sweep. That is the best that I can offer :see_no_evil:

platform/scheduler/zephyr/src/mender-scheduler.c
```diff
@@ -59,6 +89,17 @@ mender_scheduler_alt_work_create(mender_scheduler_alt_work_function_t func, int3
...
 
+#ifdef CONFIG_MENDER_SCHEDULER_SEPARATE_WORK_QUEUE
+    /* Create and start work queue */
+    k_work_queue_init(&mender_scheduler_work_queue_handle);
+    k_work_queue_start(&mender_scheduler_work_queue_handle,
+                       mender_scheduler_work_queue_stack,
+                       CONFIG_MENDER_SCHEDULER_WORK_QUEUE_STACK_SIZE * 1024,
+                       CONFIG_MENDER_SCHEDULER_WORK_QUEUE_PRIORITY,
+                       NULL);
+    k_thread_name_set(k_work_queue_thread_get(&mender_scheduler_work_queue_handle), "mender_scheduler_work_queue");
```
lluiscampos: I would make the thread name shorter. This is printed for example in the stack overflow error messages, and from what I've seen are just short ids.

What about "mender_queue" or "mender_work"?

platform/scheduler/zephyr/src/mender-scheduler.c
```diff
@@ -43,64 +43,56 @@
 #endif /* CONFIG_MENDER_SCHEDULER_WORK_QUEUE_PRIORITY */
 
 /**
- * @brief Mender scheduler work queue handle
+ * @brief Mender scheduler work queue
  */
-static struct k_work_q mender_scheduler_work_queue_handle;
+static struct k_work_q work_queue;
```
lluiscampos, line 48: Thanks - better aligns with Zephyr documentation

lluiscampos (approved):
Well done, thanks. Good to tackle the scheduler in small chunks of work!


wrt authentication interval,

> It's not respected in teh code anyway and it's unclear what its
purpose should be.

The original purpose from the community project was to have a faster interval while the device is not authenticated. As far as I know we don't have any retry mechanism, so as you intended now when the device is not authenticated it will wait for the whole poll interval to try auth again - this we need to address somehow. We don't need to do it in the current task but we have to decide how to do it and create the task soon. I vote for a retry mechanism, similar to the Linux client, with an exponential backoff time.

Or for simplicity keep the authentication interval and use it only on first boot until we get authentication :) - yes, it won't represent all the authentications action that are performed through the code as you say but it still covers this case and has a lot of value (we could then rename it to something more explicit).

wrt poll interval = 0

> But what's the point of activating Mender without it
checking for updates?

Based on KConfig comment, I think the intention was to be used in the "SDK mode", where we expose the check for deployment (and install, etc) and let the application call it at any time. (this won't play nice with our state machine / update modules design so I agree with disabling it)

platform/scheduler/zephyr/src/mender-scheduler.c
vpodzime: I'm changing this in the follow-up to this and there I name the work queue which is what we should actually do, I 
believe.

vpodzime:
I'm merging this so that I can submit the follow-up PR. Anyone tagged for a review, please feel free to do so and I'll address the comments and suggestions in the fw-up (or another one may there be a need).

PR #105, opened 2024-11-04
Scheduler rework -- part two, work items

vpodzime:
Follow-up to #104

PR #106, opened 2024-11-04
Scheduler rework -- part three, inventory and locking

vpodzime:
Based on #105 

Regarding the locking part -- to the best of my knowledge, the only part that needs locking is authentication (and `artifact_name` as @lluiscampos pointed out) which is out of scope of this PR.

### DIFF
--- a/core/src/mender-inventory.c
+++ b/core/src/mender-inventory.c
@@ -42,7 +42,7 @@ static void              *mender_inventory_mutex    = NULL;
 /**
  * @brief Mender inventory work handle
  */
-static void *mender_inventory_work_handle = NULL;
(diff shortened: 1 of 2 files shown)


vpodzime:
> Okey, but to have the locking mechanism in place is part of your current task? Or will you create a follow-up task?

There are mutexes available for both POSIX and Zephyr so all the building blocks are there. Or am I missing something?

lluiscampos:
> > Okey, but to have the locking mechanism in place is part of your current task? Or will you create a follow-up task?
> 
> There are mutexes available for both POSIX and Zephyr so all the building blocks are there. Or am I missing something?

Perfect! (I didn't check it)

PR #107, opened 2024-11-04
chore: Shorten the downloading artifact log entry

lluiscampos:
From something like:
```
[00:01:22.304,412] <inf> mender: Downloading deployment artifact with id '5f1aab23-9
629-46e1-b13e-b80033470902', artifact name 'nrf-update-inventory' and uri...
```

To something like:
```
[00:02:33.102,996] <inf> mender: Downloading artifact with id '5cb9365...', name 'nr
f-update-giving', uri 'https://c271964d41749feb10da762816c952ee.r2.cloudf...
```

So that at least part of the uri is visible.

### DIFF
--- a/core/src/mender-client.c
+++ b/core/src/mender-client.c
@@ -936,10 +936,13 @@ mender_client_update_work_function(void) {
                     goto END;
                 }
 
-                mender_log_info("Downloading deployment artifact with id '%s', artifact name '%s' and uri '%s'",
-                                deployment->id,
-                                deployment->artifact_name,
-                                deployment->uri);
+#if CONFIG_MENDER_LOG_LEVEL >= MENDER_LOG_LEVEL_INF
+                if (strlen(deployment->id) > 10) {
+                    mender_log_info("Downloading artifact with id '%.7s...', name '%s', uri '%s'", deployment->id, deployment->artifact_name, deployment->uri);
+                } else {
+                    mender_log_info("Downloading artifact with id '%s', name '%s', uri '%s'", deployment->id, deployment->artifact_name, deployment->uri);
+                }
+#endif
                 mender_client_publish_deployment_status(deployment->id, MENDER_DEPLOYMENT_STATUS_DOWNLOADING);
 
                 /* Set deployment_id */


core/src/mender-client.c
```diff
@@ -905,10 +905,16 @@ mender_client_update_work_function(void) {
...
-                                deployment->uri);
+#if CONFIG_MENDER_LOG_LEVEL >= MENDER_LOG_LEVEL_INF
+                char id_short[7];
+                if (strlen(deployment->id) >= 7) {
+                    strncpy(id_short, deployment->id, 7);
+                    strcat(id_short, "...");
+                } else {
+                    strncpy(id_short, deployment->id, strlen(deployment->id));
+                }
+                mender_log_info("Downloading artifact with id '%s', name '%s', uri '%s'", id_short, deployment->artifact_name, deployment->uri);
```
lluiscampos, line 916: It doesn't work properly, I got for example an output like `artifact with id 'a8cf648 �%...', name`. I'll look at it at a later time of maybe you C experts can spot the bug.

core/src/mender-client.c
```diff
@@ -905,10 +905,16 @@ mender_client_update_work_function(void) {
...
                 }
 
-                mender_log_info("Downloading deployment artifact with id '%s', artifact name '%s' and uri '%s'",
-                                deployment->id,
-                                deployment->artifact_name,
-                                deployment->uri);
+#if CONFIG_MENDER_LOG_LEVEL >= MENDER_LOG_LEVEL_INF
+                char id_short[7];
+                if (strlen(deployment->id) >= 7) {
+                    strncpy(id_short, deployment->id, 7);
```
vpodzime, line 911: I'd use `sizeof(id_short)` here and above so that we can simply update a single place when we decide we want this longer.

core/src/mender-client.c
vpodzime: yup, we need space for 10 characters and then use 7 from the ID + "..." or all 10 from the ID if it fits.

vpodzime: Btw, `%.7s` also works to limit the number of characters printed. So it might be easiest to do `if (strlen() <= 10) log(everything); else log(%.7s and with_dots)`.

lluiscampos: > Btw, %.7s also works to limit the number of characters printed

TIL! Thanks. I'll try this for a simpler to read code :clap:

lluiscampos:
Please also share high level thoughts about the change. Other ideas:
* remove id -> name and uri are more important
* truncate also the uri (maybe print only the host?) -> the URI is a presigned long url and we will never really print it, in practice

vpodzime (approved):
Looks good to me. And I'd keep all three in there for now. The name is the most important, but it's there just fine.

PR #108, opened 2024-11-04
chore: Force artifact_name reload on commit

lluiscampos:
The static variable `artifact_name` from the API source code is used on inventory data push and in deployments check, but it was not being refreshed from the store after a successful commit.

### DIFF
--- a/core/src/mender-api.c
+++ b/core/src/mender-api.c
@@ -66,6 +66,15 @@ static mender_err_t mender_api_http_text_callback(mender_http_client_event_t eve
 static char *artifact_name = NULL;
 
 mender_err_t
+mender_api_artifact_name_load() {
+    if ((MENDER_OK != mender_storage_get_artifact_name(&artifact_name)) && (NULL != artifact_name)) {
+        mender_log_error("Unable to get artifact name");
+        return MENDER_FAIL;
+    }
+    return MENDER_OK;
+}
+
+mender_err_t
 mender_api_init(mender_api_config_t *config) {
 
     assert(NULL != config);
@@ -74,8 +83,7 @@ mender_api_init(mender_api_config_t *config) {
     mender_err_t ret;
 
     /* Load and set artifact_name here */
-    if ((MENDER_OK != mender_storage_get_artifact_name(&artifact_name)) && (NULL != artifact_name)) {
-        mender_log_error("Unable to get artifact name");
+    if (MENDER_OK != mender_api_artifact_name_load()) {
         return MENDER_FAIL;
     }
     /* Save configuration */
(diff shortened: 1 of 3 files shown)


core/src/mender-api.c
```diff
@@ -65,6 +65,15 @@ static mender_err_t mender_api_http_text_callback(mender_http_client_event_t eve
  */
 static char *artifact_name = NULL;
```
lluiscampos, line 66: @vpodzime This variable is also shared across the two work items: check for updates and push inventory. Hence it should be protected too.

core/src/mender-client.c
```diff
@@ -485,6 +485,11 @@ mender_commit_artifact_data(void) {
         return MENDER_FAIL;
     }
 
+    if (MENDER_OK != mender_api_artifact_name_load()) {
+        /* Error already logged */
+        return MENDER_FAIL;
+    }
```
lluiscampos, line 491: All,

An alternative of doing this would be to rework both inventory and check for deployments API call to not use at all the in-memory value but always read from the persistent storage.

core/src/mender-api.c
```diff
@@ -65,6 +65,15 @@ static mender_err_t mender_api_http_text_callback(mender_http_client_event_t eve
...
 
+mender_err_t
+mender_api_artifact_name_load() {
```
vpodzime, line 69: `mender_api_` suggests it's loaded via the Mender API.

vpodzime (commented):
To me it looks like the `artifact_name` variable belongs to `mender-storage` and should be set by both `set_artifact_name()` and `get_artifact_name()` with the latter reading it from NVS if not populated already (`== NULL`).

lluiscampos:
> To me it looks like the `artifact_name` variable belongs to `mender-storage` and should be set by both `set_artifact_name()` and `get_artifact_name()` with the latter reading it from NVS if not populated already (`== NULL`).

Note that as I stated above this affects both inventory and check for deployments. I remember discussing this with @larsewi at the time of implementing deployments/next v2 but I don't remember the conclusion :see_no_evil: - why didn't we fix it then?

lluiscampos:
As @larsewi seems to be offline today, I'll take it as "do as you think it is best given today's knowledge"

I'll rework by moving the caching to the storage. layer instead

vpodzime:
> Note that as I stated above this affects both inventory and check for deployments.

Right. But that doesn't mean it cannot be stored in a `static` variable inside `mender-storage` serving basically as a write-through cache for the value in NVS.

lluiscampos:
> > Note that as I stated above this affects both inventory and check for deployments.
> 
> Right. But that doesn't mean it cannot be stored in a `static` variable inside `mender-storage` serving basically as a write-through cache for the value in NVS.

And that is exactly what I intend to do :point_up:

lluiscampos:
> I'll rework by moving the caching to the storage. layer instead

Here:
* https://github.com/mendersoftware/mender-mcu/pull/110

PR #112, opened 2024-11-05
chore: Use `const char` pointer for the Artifact name getter

lluiscampos:


larsewi (changes requested):
In `mender_storage_get_artifact_name()` there is
```C
*artifact_name = strdup("unknown");
```
if `MENDER_NOT_FOUND`. This should be changed to reference the static string to avoid memory leak. E.g.,
```C
*artifact_name = "unknown";
```

larsewi:
There does not seem to be any caching in the POSIX version. Is this the intension?

lluiscampos:
> There does not seem to be any caching in the POSIX version. Is this the intension?

To be honest I think the value of caching in Zephyr is low: we cache artifact_name which is one string (presumably to save flash reads) but then we don't cache the provides which are multiple strings.

In POSIX, specially because it is only for testing, I don't see any value in caching.

larsewi (approved):

PR #113, opened 2024-11-06
chore: Disable building of MbedTLS tests and utilities

vpodzime:
To save CPU cycles.


lluiscampos (approved):
Thank you very much for looking into this!

PR #115, opened 2024-11-07
fix: fail deployment when it's aborted

danielskinstad:
I didn't know if we wanted to fail on everything except `204`, so I added `MENDER_ABORTED`. If we can go to failure on everything except a `204`, then we could just check for `MENDER_FAIL`
I also added a check to reboot just in case you abort right after the check in install

### DIFF
--- a/core/src/mender-client.c
+++ b/core/src/mender-client.c
@@ -924,12 +924,13 @@ mender_client_update_work_function(void) {
                     mender_log_info("Downloading artifact with id '%s', name '%s', uri '%s'", deployment->id, deployment->artifact_name, deployment->uri);
                 }
 #endif
-                mender_client_publish_deployment_status(deployment->id, MENDER_DEPLOYMENT_STATUS_DOWNLOADING);
-
                 /* Set deployment_id */
                 deployment_id = deployment->id;
 
-                if (MENDER_OK == (ret = mender_download_artifact(deployment->uri, mender_client_deployment_data, &mender_update_module))) {
+                /* Check ret to see if the deployment is aborted */
+                ret = mender_client_publish_deployment_status(deployment->id, MENDER_DEPLOYMENT_STATUS_DOWNLOADING);
+                if ((MENDER_ABORTED != ret)
+                    && (MENDER_OK == (ret = mender_download_artifact(deployment->uri, mender_client_deployment_data, &mender_update_module)))) {
                     assert(NULL != mender_update_module);
 
                     /* Get artifact context if artifact download succeeded */
@@ -971,8 +972,9 @@ mender_client_update_work_function(void) {
 
             case MENDER_UPDATE_STATE_INSTALL:
                 mender_log_info("Download done, installing artifact");
-                mender_client_publish_deployment_status(deployment_id, MENDER_DEPLOYMENT_STATUS_INSTALLING);
-                if (NULL != mender_update_module->callbacks[update_state]) {
+                /* Check ret to see if the deployment is aborted */
+                ret = mender_client_publish_deployment_status(deployment_id, MENDER_DEPLOYMENT_STATUS_INSTALLING);
+                if ((MENDER_ABORTED != ret) && (NULL != mender_update_module->callbacks[update_state])) {
                     ret = mender_update_module->callbacks[update_state](update_state, (mender_update_state_data_t)NULL);
                 }
                 if ((MENDER_OK == ret) && !mender_update_module->requires_reboot) {
@@ -988,8 +990,9 @@ mender_client_update_work_function(void) {
             case MENDER_UPDATE_STATE_REBOOT:
                 assert(mender_update_module->requires_reboot);
                 mender_log_info("Artifact installation done, rebooting");
-                mender_client_publish_deployment_status(deployment_id, MENDER_DEPLOYMENT_STATUS_REBOOTING);
-                if ((MENDER_OK == ret) && (NULL != mender_update_module->callbacks[update_state])) {
+                /* Check ret to see if the deployment is aborted */
+                ret = mender_client_publish_deployment_status(deployment_id, MENDER_DEPLOYMENT_STATUS_REBOOTING);
+                if ((MENDER_ABORTED != ret) && (NULL != mender_update_module->callbacks[update_state])) {
                     /* Save the next state before running the reboot callback --
                      * if there is an interrupt (power, crash,...) right after,
                      * it will reboot anyway so after the new boot, reboot
--- a/core/src/mender-api.c
+++ b/core/src/mender-api.c
@@ -600,6 +600,10 @@ mender_api_publish_deployment_status(const char *id, mender_deployment_status_t 
     if (204 == status) {
         /* No response expected */
         ret = MENDER_OK;
+    } else if (409 == status) {
+        /* Deployment aborted */
+        mender_api_print_response_error(response, status);
+        ret = MENDER_ABORTED;
     } else {
         mender_api_print_response_error(response, status);
         ret = MENDER_FAIL;
--- a/core/src/mender-zephyr-image-update-module.c
+++ b/core/src/mender-zephyr-image-update-module.c
@@ -141,7 +141,7 @@ mender_zephyr_image_set_pending_image(MENDER_NDEBUG_UNUSED mender_update_state_t
         return MENDER_FAIL;
     }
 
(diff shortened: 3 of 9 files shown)


core/src/mender-client.c
```diff
@@ -990,7 +990,10 @@ mender_client_update_work_function(void) {
 
             case MENDER_UPDATE_STATE_INSTALL:
                 mender_log_info("Download done, installing artifact");
-                mender_client_publish_deployment_status(deployment_id, MENDER_DEPLOYMENT_STATUS_INSTALLING);
+                if (MENDER_ABORTED == mender_client_publish_deployment_status(deployment_id, MENDER_DEPLOYMENT_STATUS_INSTALLING)) {
+                    update_state = MENDER_UPDATE_STATE_FAILURE;
+                    continue;
```
vpodzime, line 995: We need to also save the state, etc., right? Let's just assign `ret` here and check it also before calling the update module's callback below so that the code skips everything until `NEXT_STATE` if `ret != MENDER_OK`.

lluiscampos (approved):

core/src/mender-client.c
```diff
@@ -943,7 +943,9 @@ mender_client_update_work_function(void) {
                     mender_log_info("Downloading artifact with id '%s', name '%s', uri '%s'", deployment->id, deployment->artifact_name, deployment->uri);
                 }
 #endif
-                mender_client_publish_deployment_status(deployment->id, MENDER_DEPLOYMENT_STATUS_DOWNLOADING);
+                if (MENDER_ABORTED == mender_client_publish_deployment_status(deployment->id, MENDER_DEPLOYMENT_STATUS_DOWNLOADING)) {
+                    ret = MENDER_FAIL;
+                }
```
vpodzime, line 948: This does nothing, basically, because there's a `ret = mender_download_artifact()` right below which overwrites this value and there's nothing in between to make it skipped.

core/src/mender-client.c
```diff
@@ -1007,7 +1011,11 @@ mender_client_update_work_function(void) {
             case MENDER_UPDATE_STATE_REBOOT:
                 assert(mender_update_module->requires_reboot);
                 mender_log_info("Artifact installation done, rebooting");
-                mender_client_publish_deployment_status(deployment_id, MENDER_DEPLOYMENT_STATUS_REBOOTING);
+                if (MENDER_ABORTED == mender_client_publish_deployment_status(deployment_id, MENDER_DEPLOYMENT_STATUS_REBOOTING)) {
+                    update_state = MENDER_UPDATE_STATE_FAILURE;
+                    set_and_store_state(update_state);
+                    continue;
```
vpodzime, line 1017: Why not leave `ret` and `NEXT_STATE` handle this just like in the `INSTALL` state?

vpodzime (changes requested):

core/src/mender-client.c
danielskinstad: Well, reboot goes to `MENDER_UPDATE_STATE_ROLLBACK` on failure, do we really want to go there?

core/src/mender-client.c
danielskinstad: Ah :sweat_smile:, misread it as `MENDER_OK != ret` for some reason, thanks

core/src/mender-client.c
vpodzime: Why not? Shouldn't we roll back the aborted deployment?

danielskinstad: We haven't rebootet yet at this stage, though. Won't going to rollback end up in calling the reboot callback, which will be wrong, since we haven't actually switched to the other partition?

vpodzime: My thinking is that an aborted deployment should be like any other failure. So here, in particular, I don't see why it should be treated differently (state-wise) from a failure of the update module's callback for the `REBOOT` state. Any exception to the success/failure state transitions must be justified as absolutely necessary. If following the "failure path" here can cause issues, we should fix the failure path, I believe.

danielskinstad: Yeah, when I think about it, it does indeed make sense to deviate as little as possible from the normal flow, thanks for your reasoning

core/src/mender-client.c
```diff
@@ -1069,6 +1081,9 @@ mender_client_update_work_function(void) {
                 if (!mender_update_module->supports_rollback) {
                     mender_log_warning("Rollback not supported for artifacts of type '%s'", mender_update_module->artifact_type);
                     ret = MENDER_FAIL;
+                } else if (aborted_deployment) {
+                    /* Don't rollback if deployment is aborted */
```
vpodzime, line 1085: Wouldn't you as a user expect a rollback if you abort a deployment that isn't committed yet?

danielskinstad: Right - but the code doesn't support an abort after a reboot - there is nothing publishing the status between reboot and commit, and on commit we've said it's too late. My idea is that before we have rebooted, we should not rollback into the other partition, but just go to failure - if not we'll end up in the wrong partition if we e.g. abort the deployment after the install state has published its status and set the pending image

danielskinstad: Assume we have `MENDER_UPDATE_STATE_DOWNLOAD -> MENDER_UPDATE_STATE_INSTALL -> MENDER_UPDATE_STATE_REBOOT`.
On each of these states we publish the status and check if the deployment is aborted. If it's aborted when we enter `MENDER_UPDATE_STATE_REBOOT`, then what do we want to do? We haven't rebooted yet as this is caught before the reboot callback is called, so we can either go directly to failure - which deviates from the normal flow, but it does make sense since we haven't actually rebooted, and this is what I originally did, or if we follow the normal state transitions we should go the failure route, which will transition to rollback, but indeed, we have nothing to rollback, which is why I checked for an aborted deployment and set `ret = MENDER_FAIL` in `MENDER_UPDATE_STATE_ROLLBACK`

If we checked if the status is aborted in `MENDER_UPDATE_STATE_COMMIT` (which Lluis said was too late, but I don't quite know why it's too late), then it would make sense to do a proper rollback, as we have rebooted into the new partition, and we would then need to perform a rollback

vpodzime: > If it's aborted when we enter `MENDER_UPDATE_STATE_REBOOT`, then what do we want to do? We haven't rebooted yet as this is caught before the reboot callback is called

That brings us to my first reaction to this -- how does it differ from the case where the reboot callback fails? Then we also haven't rebooted yet and that goes to the rollback because that's the failure path, right? I believe the `ROLLBACK` state is responsible for recognizing that it should do nothing. But not by checking `if (aborted_deployment)` but by checking if anything needs to be done to rollback. If not, it should be a no-op.

danielskinstad: Right - but as of now, it will rollback if I don't check it explicitly, so changing the functionality of rollback seems out of scope for this PR at least, I can remove the check for aborted in rollback, and then we should probably create a separate ticket for a no-op rollback?

vpodzime: Works for me! We still have this in the default update module:
```
/* no need for a rollback callback because a reboot without image confirmation is a rollback */
```
so I guess it should just work? Better double-check for sure, but my understanding is that `ROLLBACK` does nothing and then it goes to `ROLLBACK_REBOOT` which reboots the device and goes to `VERIFY_ROLLBACK_REBOOT` and then to `FAILURE`. So if the deployment abortion happens as you described above, it will simply reboot and follow the failure path. No big harm done except that the reboot could be skipped which is what we can fix/improve.

core/src/mender-client.c
```diff
@@ -971,8 +972,9 @@ mender_client_update_work_function(void) {
 
             case MENDER_UPDATE_STATE_INSTALL:
                 mender_log_info("Download done, installing artifact");
-                mender_client_publish_deployment_status(deployment_id, MENDER_DEPLOYMENT_STATUS_INSTALLING);
-                if (NULL != mender_update_module->callbacks[update_state]) {
+                /* Check ret to see if the deployment is aborted */
+                ret = mender_client_publish_deployment_status(deployment_id, MENDER_DEPLOYMENT_STATUS_INSTALLING);
+                if ((MENDER_ABORTED != ret) && NULL != mender_update_module->callbacks[update_state]) {
```
vpodzime, line 977: Missing parentheses.

vpodzime (approved):
Just one last nitpick. Thanks for your patience here! It will be interesting to see the integration tests for this :wink:  :grin:

PR #117, opened 2024-11-07
fix: allow compressed files inside a payload

danielskinstad:
* We don't care if the files _inside_ a payload are compressed, we only want to check for compressed artifacts, i.e. that the payload is not compressed.

* Moved the check to `artifact_read_manifest` to avoid having to call it for every file.

* Check the return value of `artifact_download_callback`

Changelog: Title
Ticket: MEN-7594

### DIFF
--- a/core/src/mender-artifact.c
+++ b/core/src/mender-artifact.c
@@ -329,6 +329,7 @@ mender_artifact_get_ctx(mender_artifact_ctx_t **ctx) {
 static bool
 is_compressed(const char *filename) {
 
+    /* Mender artifact supports `.gz`, `.xz`, and `.zst` */
     static const char *compression_suffixes[] = { ".gz", ".xz", ".zst", NULL };
 
     for (size_t i = 0; NULL != compression_suffixes[i]; i++) {
@@ -385,12 +386,6 @@ mender_artifact_process_data(mender_artifact_ctx_t *ctx, void *input_data, size_
 
     /* Parse data */
     do {
-
-        /* We do not support compressed artifacts */
-        if (is_compressed(ctx->file.name)) {
-            mender_log_error("Artifact compression is not supported");
-            return MENDER_FAIL;
-        }
 
         /* Treatment depending of the stream state */
         if (MENDER_ARTIFACT_STREAM_STATE_PARSING_HEADER == ctx->stream_state) {
@@ -744,6 +739,12 @@ artifact_read_manifest(mender_artifact_ctx_t *ctx) {
 
         const char *checksum_str = line;
         const char *filename     = separator + 2;
+
+        /* We do not support compressed artifacts */
+        if (mender_utils_strbeginswith(filename, "header.tar") && is_compressed(filename)) {
+            mender_log_error("Artifact compression is not supported");
+            return MENDER_FAIL;
+        }
 
         /* Useful when debugging artifact integrity check failures */
         mender_log_debug("%s  %s", checksum_str, filename);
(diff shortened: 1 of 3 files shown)


core/src/mender-artifact.c
```diff
@@ -696,6 +690,12 @@ artifact_read_manifest(mender_artifact_ctx_t *ctx) {
         const char *checksum_str = line;
         const char *filename     = separator + 2;
 
+        /* We do not support compressed artifacts */
+        if (mender_utils_strbeginswith(filename, "header.tar") && is_compressed(filename)) {
+            mender_log_error("Artifact compression is not supported");
+            return MENDER_FAIL;
+        }
```
lluiscampos, line 697: We need to check also for the payload file `data/0000.tar` (not the contents inside, but the payload which is what we treat it cannot be compressed neither)

danielskinstad: Can the payload file be compressed without the header being compressed? How can I test this, i.e. create an artifact with `-compression none` where the payload file is compressed?

core/src/mender-artifact.c
```diff
@@ -1152,6 +1152,20 @@ artifact_read_data_prepare(mender_artifact_ctx_t *ctx, mender_artifact_download_
         return MENDER_FAIL;
     }
 
+    size_t file_name_length = strlen(ctx->file.name);
+    if ((strlen("data/xxxx.tar.xx") == file_name_length) || (strlen("data/xxxx.tar.xxx") == file_name_length)) {
```
vpodzime, line 1156: What about things like `.tgz` and similar?

core/src/mender-artifact.c
```diff
@@ -1152,6 +1152,20 @@ artifact_read_data_prepare(mender_artifact_ctx_t *ctx, mender_artifact_download_
...
+        /*
+        * We allow compressed files _inside_ a payload:
+        *   'data/0000.tar/compressed.tar.gz'
+        * But not a compressed payload:
+        *   'data/0000.tar[.gz|.xz|.zst]'
```
vpodzime, line 1161: I feel like we should implement this logic in a better way than just checking lengths of strings.

platform/net/zephyr/src/mender-http.c
```diff
@@ -304,6 +304,11 @@ mender_http_artifact_download(const char *uri, mender_artifact_download_data_t *
         goto END;
     }
 
+    /* Artifact download failed */
+    if (MENDER_OK != dl_data->ret) {
+        goto END;
+    }
```
vpodzime, line 310: Please check if we need to change this in the curl-based implementation as well.

core/src/mender-artifact.c
danielskinstad: The artifact [documentation](https://github.com/mendersoftware/mender-artifact/blob/master/Documentation/artifact-format-v3.md#compression) only state that `xz`, `zt`, and `zst` are supported.

core/src/mender-artifact.c
danielskinstad: We do also check if it's compressed, though. Asserting the length is just a way of ensuring that we don't also catch files like `data/0000.tar/compressed.tar.gz`. The files will start with e.g. `data/000.tar.` either way, so shouldn't checking if the size is `data/0000.tar[.gz|.xz|.zst]`, and then checking the suffix be sufficient? Do you perhaps have another suggestion

core/src/mender-artifact.c
vpodzime: Worth adding a comment then.

core/src/mender-artifact.c
vpodzime: I think it's fine, but there should be comment explaining why checking the length only is good enough in this particular place.

danielskinstad: Just want to make sure that you know that it is also checked  with `is_compressed` below the comment, we don't only check the length - we also check the suffix of the filename

lluiscampos (approved):

PR #118, opened 2024-11-07
chore: Avoid repetitive meta data gathering for data being processed

vpodzime:
Instead of determining the values of various meta data items for the data being processed over and over with every chunk (512 bytes) of the data, let's just do it once for every file and reuse the information for every chunk in the file. I.e. once instead of 800 times for a 400KiB artifact data file.

Ticket: MEN-7645
Changelog: none


larsewi (approved):
Great :rocket:

PR #119, opened 2024-11-19
chore: Implicit (re-)authentication

vpodzime:
Multiple changes that make sure that:
- authentication happens when the first API request needs it
- getting `401 Unauthorized` results in the saved JWT token to be thrown away
- `401 Unauthorized` results in a (re-)authentication attempt and a retry of the rejected API request
- the manipulation with the JWT token is protected by a lock so that concurrent threads/tasks don't step on each other's toes when using it


core/src/mender-api.c
```diff
@@ -446,6 +457,11 @@ mender_api_publish_deployment_status(const char *id, mender_deployment_status_t
     char        *response     = NULL;
     int          status       = 0;
 
+    if (MENDER_OK != (ret = mender_api_ensure_authenticated())) {
```
larsewi: What if it returns `MENDER_DONE` (i.e., already authenticated)? __NB:__ _There are also some other places I saw this._

core/src/mender-api.c
```diff
@@ -214,6 +214,38 @@ mender_api_perform_authentication(void) {
     return ret;
 }
 
+/**
+ * @see mender_http_perform()
+ */
+static mender_err_t
+authenticated_http_perform(char                *path,
```
larsewi: I'm guessing some of these arguments can be `const` pointers

larsewi (approved):
Great stuff :rocket: only a few minor comments

core/src/mender-api.c
vpodzime: You'd think so. But they are passed to `mender_http_perform()` which doesn't take them as `const` (because they are passed to other functions missing `const`). So yes, they should be `const`, but that's a bit out of scope.

core/src/mender-api.c
vpodzime: Good catch! I wonder if I should change the function to just return `MENDER_OK` in all successful cases. Or maybe we could have a macro checking for errors vs. good `mender_err_t` values?

core/src/mender-api.c
```diff
@@ -214,6 +214,38 @@ mender_api_perform_authentication(void) {
...
+        return ret;
+    }
+    assert(NULL != api_jwt);
+
+    if (MENDER_OK != (ret = mender_http_perform(api_jwt, path, method, payload, signature, callback, params, status))) {
+        /* HTTP errors already logged. */
+        return ret;
+    }
+
+    if (401 == *status) {
```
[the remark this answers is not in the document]
vpodzime: Yeah. And also, the [Mender API docs](https://docs.mender.io/api/#device-api-deployments-report-deployment-log) list those codes as numbers so it's nice to be able to cross-check they are the same.
image

core/src/mender-api.c
larsewi: > I wonder if I should change the function to just return MENDER_OK in all successful cases.

Yepp, can't think of any immediate reason for why the caller should really care about whether it already was authenticated or not

include/mender-utils.h
```diff
@@ -95,6 +95,8 @@ typedef enum {
     MENDER_LOCK_FAILED     = -5, /**< Locking failed */
 } mender_err_t;
 
+#define MENDER_IS_ERROR(err_t_ret) ((err_t_ret) < 0)
```
larsewi, line 98: :+1:

core/src/mender-api.c
```diff
@@ -90,38 +95,39 @@ mender_api_is_authenticated(void) {
...
         mender_log_error("Unable to get public key");
         goto END;
     }
 
-    /* Get identity */
-    if (MENDER_OK != (ret = api_config.identity_cb(&identity))) {
-        mender_log_error("Unable to get identity");
-        goto END;
-    }
+    if (NULL == identity_info) {
```
lluiscampos, line 171: As we discussed in the past, the caching on the identity data is a behavior difference from the classic client. I understand the purpose, but this may hide problems during an update. Example:

An update module faulty implements something in the install state where the identity would change. This module does not require reboot. The update will continue and successfully commit. On next boot the device will be unauthenticated.

include/mender-client.h
```diff
@@ -64,7 +64,7 @@ typedef struct {
     mender_err_t (*network_release)(void);                                 /**< Invoked when mender-client releases access to the network */
     mender_err_t (*deployment_status)(mender_deployment_status_t, char *); /**< Invoked on transition changes to inform of the new deployment status */
     mender_err_t (*restart)(void);                                         /**< Invoked to restart the device */
-    mender_err_t (*get_identity)(mender_identity_t **identity);            /**< Invoked to retrieve identity */
+    mender_err_t (*get_identity)(const mender_identity_t **identity);      /**< Invoked to retrieve identity */
```
lluiscampos, line 67: This change is good. Let's be aware that it will break `mender-mcu-integration` build (right?).

lluiscampos (approved):
Splendid. It is a pleasure to go through these commit by commit :partying_face: 

See comment below - I think we should drop the cache, but YMMV.

include/mender-client.h
vpodzime: yup, I'll adapt that.

core/src/mender-api.c
vpodzime: > The update will continue and successfully commit.

This will happen even if we only cache authentication info, right? Because the JWT token will still be valid and the client will still be "authenticated".

Not caching authentication info would be quite foolish, of course. But should we drop the JWT token when an update doesn't require reboot so that it re-authenticates as part of its verification?

lluiscampos: > This will happen even if we only cache authentication info, right? Because the JWT token will still be valid and the client will still be "authenticated".

But this is also wrong on its own. Before commit we have to force a re-authentication for this very same reason: not to commit and then run into unauthenticated state. See [here](https://github.com/mendersoftware/mender/blob/4.0.4/src/mender-update/daemon/states.cpp#L669).

> But should we drop the JWT token when an update doesn't require reboot so that it re-authenticates as part of its verification?

Exactly. Although I would do it from the commit state code regardless of requiring a reboot or not.

include/mender-client.h
vpodzime: Alright, you adapt it then. :wink:

core/src/mender-api.c
vpodzime: > Although I would do it from the commit state code regardless of requiring a reboot or not.

The commit state commits the update (i.e. does `mender_flash_confirm_image()` unconditionally), there's no check. The authentication check we have is in the `VERIFY_REBOOT` state which is only entered for updates that require reboot. For those that don't, `INSTALL` goes directly to `COMMIT` with no check.

vpodzime: I added a function to trash the JWT. But I still don't understand where we should call it and where we should check that we can re-authenticate.

lluiscampos: I looked now at the code now,

> The commit state commits the update (i.e. does mender_flash_confirm_image() unconditionally), there's no check. 

Here with this reasoning you are talking about the implementation details of `zephyr-image` module, not our state machine.

What I am suggesting is to add `mender_api_drop_authentication_data()` and `mender_api_ensure_authenticated()` to the state machine code, inside `MENDER_UPDATE_STATE_COMMIT` state. If it fails here, it will go into rollback and eventually failure path.

What problems do you see doing it this way?

This is the closest to the design of the Mender C++ client, where the explicit re-auth is done in the state machine, not in the update module. This is important because we don't control the implementation of the update module.

> The authentication check we have is in the VERIFY_REBOOT state which is only entered for updates that require reboot. For those that don't, INSTALL goes directly to COMMIT with no check.

This is how our `zephyr-image` module is designed, but other modules may not care about being authenticated on `VERIFY_REBOOT` or have other requirements to check that the reboot was successful.

vpodzime: > What I am suggesting is to add `mender_api_drop_authentication_data()` and `mender_api_ensure_authenticated()` to the state machine code, inside `MENDER_UPDATE_STATE_COMMIT` state. If it fails here, it will go into rollback and eventually failure path.
> 
> What problems do you see doing it this way?

Well, nothing apart from my generic feeling that authentication should not be the measure of an update being good or not. I can totally imagine cases where the server (or network) is not available all the time and rejecting an update just because there was, for example, a short network outage after rebooting into it doesn't sound right. It can be a totally valid approach, but isn't that a decision to be made by the update module developer? I'll add a commit implementing this, but my gut feeling is that it should be possible to disable at build time, at least.

lluiscampos: The explicit re-authentication in the classic Mender client is a painful lesson based on real worlds cases. Not having this catch can brick a device. To fail an update unnecessarily is preferable to commit a potentially broken update. And to determine that an update is valid we take the connectivity to the Mender server as a requirement.

lluiscampos: Okey, optional sounds good :memo:

lluiscampos (approved):
Thank you

PR #122, opened 2024-11-25
fix: Make the mapping in mender_utils_http_status_to_string() static

vpodzime:
So that it doesn't end up on the stack.

Also change the type of the `str` field to `const char *` because it always points to string literals.


larsewi (approved):

danielskinstad (approved):
Nice, we can probably remove a lot of the unused status-strings as well, but that might as well be done in another PR

vpodzime:
> This is just my personal opinion

I think we all agree with that. I just wanted a quick fix for the most obvious and worst. :) Feel free to follow-up.

PR #123, opened 2024-11-26
Error/hang detection

vpodzime:
Detection of network errors and an attempt to fix them by network reset plus detection of being stuck waiting for a reboot and getting away from it by failing the deployment and resuming normal operation.


core/src/mender-error-counters.c
```diff
@@ -0,0 +1,55 @@
...
+#include "mender-log.h"
+#include "mender-utils.h"
+
+#define NETWORK_ERR_LIMIT 30
+
+static uint8_t net_errors = 0;
+
+mender_err_t
+mender_err_count_net_inc(void) {
+    if (net_errors < UINT8_MAX) {
```
lluiscampos, line 32: This is okey as long as `#define NETWORK_ERR_LIMIT` is in this same file. But if we later move it around we risk having `NETWORK_ERR_LIMIT > 256` and then the error detection will never trigger.

(will the compiler throw a warning below for `net_errors > NETWORK_ERR_LIMIT` if NETWORK_ERR_LIMIT overflows `uint8_t`?)

lluiscampos (commented):
>     * We could add locking, but I feel like it's not worth it because if we miss an error because of interleaving threads, it's not a big deal.
>     * We could make the code platform-specific and use atomic operations, but, again, not worth it for the same reason.

Both "tasks" run in the same thread (through the same workqueue), we should be safe in this aspect.

>     * Network issues are a good first example, what else should we try to detect?

I don't know...

>     * Should we return `MENDER_OK` to the scheduler while waiting for a reboot so that we have a (better) chance to detect that the reboot is not happening and we should fail the deployment? (introducing a `pending_reboot` counter)

Something along these lines, I think. Returning MENDER_DONE means no further re-schedule, right? So from there on we don't have any chance to detect these problems.

>     * Should the limits be build-configurable?

If we take better care of the type casting of the limit (ref comment below) then I would say yes because it is cheap and opens up for more user cases. We don't have user feecback yet, but I can imagine systems where the connections are expected to fail and maybe they want like > 100 error tolerance but also the opposite systems where a single failure is already enough to cancel an update :shrug: 

>     * Should this functionality be optional?

The motivation being for saving ROM space? I would not leave it optional, as we believe this is a safety mechanism to have for robust updates, but still make the limit configurable so that can be set to 1 (see above)

> Thanks in advance for feedback!

You are welcome :beers:

core/src/mender-error-counters.c
vpodzime: > will the compiler throw a warning below for `net_errors > NETWORK_ERR_LIMIT` if NETWORK_ERR_LIMIT overflows `uint8_t`

I don't think so. But we have multiple ways to make sure the limit is not above 255, right? Kconfig supports range, we can do `static_assert` or even 
```
#if NETWORK_ERR_LIMIT > UINT8_MAX
#error NETWORK_ERR_LIMIT cannot be greater than UINT8_MAX
#endif
```

danielskinstad (approved):
Nice, I haven't tested it because of the current problems with downloading artifacts, but it looks good to me

vpodzime:
> Nice, I haven't tested it because of the current problems with downloading artifacts

Same here :sweat:

lluiscampos (approved):

lluiscampos:
> Nice, I haven't tested it because of the current problems with downloading artifacts, (...)

@danielskinstad We have [a fix for the Artifacts download](https://github.com/mendersoftware/mender-mcu-integration/pull/31) now, can you then spend some time testing this PR? (Vratislav has different set of problems downloading Artifacts)

vpodzime:
> deduped and instead instead have an enum for which counter to operate on? Might save some static memory?

It might, but also might not, we would have to check that. But right now the `iram` segment is not a problem for us on any of the board, AFAIK, so that wouldn't help us much. And the code would be quite a bit harder to read with all the `ifdef`s.

larsewi (approved):

PR #124, opened 2024-11-29
Artifact unit tests

danielskinstad:


core/src/mender-artifact.c
```diff
@@ -1224,10 +1224,11 @@ artifact_read_data_prepare(mender_artifact_ctx_t *ctx, mender_artifact_download_
         mender_log_error("Unable to allocate memory");
         return MENDER_FAIL;
     }
-    for (char *ch = strstr(mdata_cache->checksum_fname, ".tar"); (NULL != ch) && (*ch != '\0'); ch++) {
+    char *ch = mdata_cache->checksum_fname;
+    while (NULL != (ch = strstr(ch, ".tar"))) {
         /* Don't worry! The call to strlen() on a static string should
          * be optimized out by the compiler */
-        *ch = ch[strlen(".tar")];
+        memmove(ch, ch + strlen(".tar"), strlen(ch + strlen(".tar")) + 1);
```
danielskinstad, line 1235: This was the only way I could get the address sanitizer to pass, the simple pointer arithmetic kept throwing heap buffer overflow errors, and I couldn't find another way to work around it

lluiscampos: It looks good to me but I am looking forward to hear from our C experts too :monocle_face:

tests/unit/core/mender-artifact_test.cpp
```diff
@@ -0,0 +1,229 @@
...
+    }
+
+    mender_update_module_t          *update_module;
+    mender_deployment_data_t        *deployment_data;
+    mender_artifact_download_data_t *mock_download_data;
+
+    void SetUp() {
+        update_module      = (mender_update_module_t *)malloc(sizeof(mender_update_module_t));
+        deployment_data    = (mender_deployment_data_t *)malloc(sizeof(mender_deployment_data_t));
+        mock_download_data = (mender_artifact_download_data_t *)malloc(sizeof(mender_artifact_download_data_t));
```
lluiscampos, line 91: On first read I thought that it didn't make sense to use dynamic memory for these on the tests, and we should just use C++ (implicit) constructor/destructor on these `struct`. But then I saw that `mender_update_module_unregister_all` does free the `update_module`... From the API point of vies, is it really correct that `mender_update_module_unregister_all` takes ownership of the `update_module`?

tests/unit/core/mender-artifact_test.cpp
```diff
@@ -0,0 +1,229 @@
...
+        create_download_data(update_module, deployment_data, mock_download_data);
```
lluiscampos, line 92: This function does more than just "craeate data". Is more about initializing the client or the module.

Consider renaming.

tests/unit/core/mender-artifact_test.cpp
```diff
@@ -0,0 +1,229 @@
...
+    mender_artifact_release_ctx(ctx);
+}
+
+TEST_F(MenderArtifactTest, GetCtx) {
+    mender_artifact_ctx_t *ctx = mender_artifact_create_ctx(64);
+
+    mender_artifact_ctx_t *ctx2;
+    if (MENDER_OK != mender_artifact_get_ctx(&ctx2)) {
+        FAIL();
+    }
```
lluiscampos, line 120: IMO, it would read better like:
```
ASSERT_EQ(mender_artifact_get_ctx(&ctx2), MENDER_OK);
```

lluiscampos: There are few of these below, I won't comment on each :eyes:

lluiscampos (commented):
Difficult to argue with this :point_down: 

image


Well done! :chart_with_upwards_trend: 

However, looking at the coverage report, we can still do a bit more:
* `is_compressed` with actual compressed filenames
    * Additionally, in the past we talked about moving this one and possibly others into utils, have you considered that?
* `mender_artifact_get_ctx` with no existing context
* artifact with meta-data
* This part:
image
* And this one (I would have expected to be covered with the current tests :thinking: 
image

core/src/mender-artifact.c
vpodzime: I'm a bit hesitant to use sub-optimal code only to make a (buggy, right?) sanitizer happy. What exactly did it not like? And did it give any details?

tests/unit/core/mender-artifact_test.cpp
```diff
@@ -0,0 +1,314 @@
...
+
+    /* The original context size was too small, so the new one should be larger
+     * than the original isze */
+    ASSERT_GT(ctx->input.size, ctx->input.orig_size);
+
+    /* Should be 0/NULL, as there are no artifacts currently
+     * being parsed */
+    ASSERT_EQ(ctx->file.size, 0);
+    ASSERT_EQ(ctx->file.index, 0);
+    ASSERT_STREQ(ctx->file.name, nullptr);
```
vpodzime, line 175: `ASSERT_EQ(ctx->file.name, NULL)`

tests/unit/core/mender-artifact_test.cpp
```diff
@@ -0,0 +1,314 @@
...
+    /* Check the stream state */
+    ASSERT_EQ(ctx->stream_state, MENDER_ARTIFACT_STREAM_STATE_PARSING_HEADER);
+
+    mender_artifact_release_ctx(ctx);
+}
+
+TEST_F(MenderArtifactTest, ProcessData_TooLargeSize) {
+    mender_artifact_ctx_t *ctx = mender_artifact_create_ctx(1024);
+
+    /* Assert that if file.size is too big, we realloc correctly */
```
vpodzime, line 194: ? This deserves a longer comment, I think.

tests/unit/core/mender-artifact_test.cpp
```diff
@@ -0,0 +1,314 @@
...
+    ctx->input.length = 99999999999999999;
+    ASSERT_EQ(MENDER_FAIL, mender_artifact_process_data(ctx, data, artifact_data.size(), mock_download_data));
+
+    mender_artifact_release_ctx(ctx);
+}
+
+TEST_F(MenderArtifactTest, ProcessData_EmptyInput) {
+    mender_artifact_ctx_t *ctx = mender_artifact_create_ctx(1024);
+
+    ASSERT_EQ(MENDER_OK, mender_artifact_process_data(ctx, nullptr, 0, mock_download_data));
```
vpodzime, line 212: Please use `NULL` when calling C functions.

tests/unit/core/mender-artifact_test.cpp
```diff
@@ -0,0 +1,314 @@
...
+
+    ASSERT_EQ(ctx->input.size, ctx->input.orig_size);
+    ASSERT_EQ(ctx->file.size, 0);
+    ASSERT_EQ(ctx->file.index, 0);
+    ASSERT_STREQ(ctx->file.name, nullptr);
```
vpodzime, line 217: `ASSERT_EQ(..., NULL)`

tests/unit/core/mender-artifact_test.cpp
```diff
@@ -0,0 +1,314 @@
...
+        DESTROY_AND_NULL(mender_delete_deployment_data, mock_download_data->deployment);
+        free(deployment_data);
+        free(mock_download_data);
+        mender_update_module_unregister_all();
+    }
+};
+
+TEST_F(MenderArtifactTest, CreateCtx) {
+    mender_artifact_ctx_t *ctx = mender_artifact_create_ctx(64);
+    ASSERT_NE(ctx, nullptr);
```
vpodzime, line 134: The `ASSERT_*` macros stop the test. They should only be used for things that prevent other part of the test from evaluation. The rest should use `EXPECT_*`.

tests/unit/core/mender-artifact_test.cpp
```diff
@@ -0,0 +1,314 @@
...
+    ASSERT_STREQ(ctx->file.name, nullptr);
+
+    mender_artifact_release_ctx(ctx);
+}
+
+TEST_F(MenderArtifactTest, ProcessData_LargeInput) {
+    mender_artifact_ctx_t *ctx = mender_artifact_create_ctx(1024);
+    string                 large_data(10240, 'A');
+
+    ASSERT_EQ(MENDER_FAIL, mender_artifact_process_data(ctx, (void *)large_data.c_str(), large_data.size(), mock_download_data));
```
vpodzime, line 226: Does this fail because of large data or because it's not valid artifact data?

tests/unit/core/mender-artifact_test.cpp
```diff
@@ -0,0 +1,314 @@
...
+
+    mender_artifact_release_ctx(ctx);
+}
+
+TEST_F(MenderArtifactTest, ProcessData_InvalidDownloadData) {
+    mender_artifact_ctx_t          *ctx          = mender_artifact_create_ctx(1024);
+    mender_artifact_download_data_t invalid_data = {};
```
vpodzime, line 233: What is this `{}` initialization doing? Let's do `{0}` if that's what we want.

vpodzime (commented):
Looks good to me otherwise.

tests/unit/core/mender-artifact_test.cpp
danielskinstad: Yeah, thanks for noticing it

tests/unit/core/mender-artifact_test.cpp
```diff
@@ -0,0 +1,314 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <gtest/gtest.h>
+
+#include <filesystem>
+#include <fstream>
+
+extern "C" {
```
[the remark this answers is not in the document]
danielskinstad: Yeah, it's not needed :+1:

tests/unit/core/mender-artifact_test.cpp
danielskinstad: The name is indeed misleading; it has nothing to do with the size of the data

core/src/mender-artifact.c
```diff
@@ -504,9 +504,13 @@ mender_artifact_release_ctx(mender_artifact_ctx_t *ctx) {
...
         mender_utils_key_value_list_free(ctx->artifact_info.provides);
         mender_utils_key_value_list_free(ctx->artifact_info.depends);
-        for (mender_artifact_checksum_t *checksum = ctx->artifact_info.checksums; NULL != checksum; checksum = checksum->next) {
-            free(checksum->filename);
+        mender_artifact_checksum_t *checksum = ctx->artifact_info.checksums;
+        while (NULL != checksum) {
+            mender_artifact_checksum_t *next = checksum->next;
+            FREE_AND_NULL(checksum->filename);
             mender_sha256_finish(checksum->context, NULL);
+            FREE_AND_NULL(checksum);
```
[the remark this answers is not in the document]
danielskinstad: Yeah, the while-loop stuck after a debugging session, a for-loop is cleaner :+1:

tests/unit/core/mender-artifact_test.cpp
danielskinstad: Fair enough, I guess I can do something like `mender_artifact_download_data_t empty_data = {NULL, NULL, NULL, (mender_err_t)0};`, then; just setting it do {0} won't work - it throws `missing initializer for member...`

vpodzime: > just setting it do {0} won't work

Strange, it should be the same as the 4 `NULL`/`0` values. But at least it's explicit.

core/src/mender-artifact.c
vpodzime: https://github.com/mendersoftware/mender-mcu/pull/125

tests/unit/core/mender-artifact_test.cpp
lluiscampos: As the code is today, yes: it is the right way. I won't insist this late in the API design, it was just thinking out loud.

lluiscampos (approved):
There are still few more things that we could test. The most important of them is `mender_artifact_get_device_type` which is part of the API and 0% tested.

Anyway this is a good base and the coverage inside the artifact parsing code is quite alright already. Approving.

core/src/mender-artifact.c
```diff
@@ -504,10 +504,14 @@ mender_artifact_release_ctx(mender_artifact_ctx_t *ctx) {
 #ifdef CONFIG_MENDER_FULL_PARSE_ARTIFACT
         mender_utils_key_value_list_free(ctx->artifact_info.provides);
         mender_utils_key_value_list_free(ctx->artifact_info.depends);
-        for (mender_artifact_checksum_t *checksum = ctx->artifact_info.checksums; NULL != checksum; checksum = checksum->next) {
-            free(checksum->filename);
+        mender_artifact_checksum_t *next;
+        for (mender_artifact_checksum_t *checksum = ctx->artifact_info.checksums; NULL != checksum; checksum = next) {
+            FREE_AND_NULL(checksum->filename);
```
vpodzime, line 509: This can do `free(checksum->filename)`, no point in setting some value to `NULL` when we are destroying it 4 lines below. But not a biggie.

tests/unit/core/mender-artifact_test.cpp
```diff
@@ -0,0 +1,320 @@
...
+
+        os << script;
+        os.close();
+
+        chmod(script_file_name.c_str(), S_IRUSR | S_IWUSR | S_IXUSR);
+        system(script_file_name.c_str());
+
+        artifact_path = tmp_dir / "unit-test-artifact.mender";
+        ifstream file(artifact_path);
+        return vector<uint8_t>((istreambuf_iterator<char>(file)), istreambuf_iterator<char>());
```
vpodzime, line 89: Why `char`?

tests/unit/core/mender-artifact_test.cpp
```diff
@@ -0,0 +1,320 @@
...
+    }
+
+    mender_update_module_t          *update_module;
+    mender_deployment_data_t        *deployment_data;
+    mender_artifact_download_data_t *mock_download_data;
+
+    void SetUp() {
+        update_module      = (mender_update_module_t *)malloc(sizeof(mender_update_module_t));
+        deployment_data    = (mender_deployment_data_t *)malloc(sizeof(mender_deployment_data_t));
+        mock_download_data = (mender_artifact_download_data_t *)malloc(sizeof(mender_artifact_download_data_t));
```
vpodzime, line 111: The type-casts are not needed, the assignment takes care of them. But they don't hurt, of course. :)

tests/unit/core/mender-artifact_test.cpp
```diff
@@ -0,0 +1,320 @@
...
+        DESTROY_AND_NULL(mender_delete_deployment_data, mock_download_data->deployment);
+        free(deployment_data);
+        free(mock_download_data);
+        mender_update_module_unregister_all();
+    }
+};
+
+TEST_F(MenderArtifactTest, CreateCtx) {
+    mender_artifact_ctx_t *ctx = mender_artifact_create_ctx(64);
+    EXPECT_NE(ctx, nullptr);
```
vpodzime, line 128: I'd prefer `NULL` here.

tests/unit/core/mender-artifact_test.cpp
```diff
@@ -0,0 +1,320 @@
...
+
+    /* The original context size was too small, so the new one should be larger
+     * than the original isze */
+    EXPECT_GT(ctx->input.size, ctx->input.orig_size);
+
+    /* Should be 0/NULL, as there are no artifacts currently
+     * being parsed */
+    EXPECT_EQ(ctx->file.size, 0);
+    EXPECT_EQ(ctx->file.index, 0);
+    EXPECT_STREQ(ctx->file.name, NULL);
```
vpodzime, line 169: `EXPECT_EQ()`

tests/unit/core/mender-artifact_test.cpp
```diff
@@ -0,0 +1,320 @@
...
+
+TEST_F(MenderArtifactTest, ProcessData_EmptyInput) {
+    mender_artifact_ctx_t *ctx = mender_artifact_create_ctx(1024);
+
+    EXPECT_EQ(MENDER_OK, mender_artifact_process_data(ctx, NULL, 0, mock_download_data));
+
+    EXPECT_EQ(ctx->input.size, ctx->input.orig_size);
+    EXPECT_EQ(ctx->file.size, 0);
+    EXPECT_EQ(ctx->file.index, 0);
+    EXPECT_STREQ(ctx->file.name, NULL);
```
vpodzime, line 214: `EXPECT_EQ()`

tests/unit/core/mender-artifact_test.cpp
```diff
@@ -0,0 +1,320 @@
...
+
+    mender_artifact_release_ctx(ctx);
+}
+
+TEST_F(MenderArtifactTest, ProcessData_InvalidData) {
+    mender_artifact_ctx_t *ctx = mender_artifact_create_ctx(1024);
+    string                 invalid_data(1240, 'A');
```
vpodzime, line 221: I'd prefer `vector<uint8_t> invalid_data(1240, 'A')` and then using `invalid_data.data()`.

vpodzime (approved):
Just nitpicks. Feel free to cover them in a follow-up or even not at all.

tests/unit/core/mender-artifact_test.cpp
danielskinstad: The compiler, for some reason, doesn't accept that

tests/unit/core/mender-artifact_test.cpp
danielskinstad: Same as above

tests/unit/core/mender-artifact_test.cpp
danielskinstad: As far as I can tell because the iterator reads chars from the ifstream `file`

PR #125, opened 2024-12-04
fix: Fix buffer overflow when removing ".tar" from a checksum file name

vpodzime:
Checking `*ch != '\0'` in the loop condition doesn't work because by the time we do it `ch` is one byte/character futher because of the for-loop increment.

Ticket: none
Changelog: none

### DIFF
--- a/core/src/mender-artifact.c
+++ b/core/src/mender-artifact.c
@@ -1224,10 +1224,11 @@ artifact_read_data_prepare(mender_artifact_ctx_t *ctx, mender_artifact_download_
         mender_log_error("Unable to allocate memory");
         return MENDER_FAIL;
     }
-    for (char *ch = strstr(mdata_cache->checksum_fname, ".tar"); (NULL != ch) && (*ch != '\0'); ch++) {
+    bool done = false;
+    for (char *ch = strstr(mdata_cache->checksum_fname, ".tar"); (NULL != ch) && !done; ch++) {
         /* Don't worry! The call to strlen() on a static string should
          * be optimized out by the compiler */
-        *ch = ch[strlen(".tar")];
+        done = (*ch = ch[strlen(".tar")]) == '\0';
     }
 
     mdata_cache->valid = true;


core/src/mender-artifact.c
```diff
@@ -1224,10 +1224,11 @@ artifact_read_data_prepare(mender_artifact_ctx_t *ctx, mender_artifact_download_
         mender_log_error("Unable to allocate memory");
         return MENDER_FAIL;
     }
-    for (char *ch = strstr(mdata_cache->checksum_fname, ".tar"); (NULL != ch) && (*ch != '\0'); ch++) {
+    bool done = false;
+    for (char *ch = strstr(mdata_cache->checksum_fname, ".tar"); (NULL != ch) && !done; ch++) {
         /* Don't worry! The call to strlen() on a static string should
          * be optimized out by the compiler */
-        *ch = ch[strlen(".tar")];
+        done = (*ch = ch[strlen(".tar")]) != '\0';
```
danielskinstad, line 1231: Shouldn't done be set to true if it equals the null terminator, not the other way around? I.e. `== '\0` instead of `!= '\0`?

lluiscampos (commented):
> Checking *ch != '\0' in the loop condition doesn't work because by the time we do it ch is one byte/character futher because of the for-loop increment

My two cents: isn't it better to change in the for loop statement to `++ch` ?

core/src/mender-artifact.c
larsewi: Yup https://godbolt.org/z/e6x4cEGfd

vpodzime:
> My two cents: isn't it better to change in the for loop statement to `++ch` ?

You mean doing `++ch` instead of `ch++` in the for-loops increment? That makes no difference (here), AFAICT.

core/src/mender-artifact.c
vpodzime: > Shouldn't done be set to true if it equals the null terminator, not the other way around? I.e. `== '\0` instead of `!= '\0`?

Of course! Late-night's commits. :blush:

larsewi (approved):
:rocket:

danielskinstad (approved):

PR #126, opened 2025-01-06
Fix: header.tar checksum calculation in chunks

vpodzime:
Tested with
```
diff --git a/core/src/mender-artifact.c b/core/src/mender-artifact.c
index c46920c..7c3ae46 100644
--- a/core/src/mender-artifact.c
+++ b/core/src/mender-artifact.c
@@ -304,6 +304,7 @@ mender_artifact_create_ctx(size_t buf_size) {
         mender_log_error("Unable to allocate memory for artifact context");
         return NULL;
     }
+    mender_log_info("Starting with a buffer size of %zd", buf_size);
     if (NULL == (ctx->input.data = malloc(buf_size))) {
         mender_log_error("Unable to allocate memory for artifact context buffer");
         free(ctx);
@@ -360,6 +361,7 @@ mender_artifact_process_data(mender_artifact_ctx_t *ctx, void *input_data, size_
     /* Copy data to the end of the internal buffer */
     if ((NULL != input_data) && (0 != input_length)) {
         if ((ctx->input.length + input_length) > ctx->input.size) {
+            mender_log_error("Current buffer size %zd is too small for %zd", ctx->input.size, (ctx->input.length + input_length));
             new_size = ctx->input.length + input_length;
             /* data/ files are processed per block for which the original size of the buffer should
                be enough, but metadata is processed as whole files so there we expect we will need
```
showing
```
[00:03:00.660,000] <inf> mender: Starting with a buffer size of 1536
```
and nothing else. :partying_face:

### DIFF
--- a/core/src/mender-artifact.c
+++ b/core/src/mender-artifact.c
@@ -363,8 +363,11 @@ mender_artifact_process_data(mender_artifact_ctx_t *ctx, void *input_data, size_
             new_size = ctx->input.length + input_length;
             /* data/ files are processed per block for which the original size of the buffer should
                be enough, but metadata is processed as whole files so there we expect we will need
-               more. */
+               more, except for header.tar (and tarballs in general) which are processed
+               transparently. */
             if (mender_utils_strbeginswith(ctx->file.name, "data/")) {
+                expected_required = ctx->input.orig_size;
+            } else if (mender_utils_strendswith(ctx->file.name, ".tar")) {
                 expected_required = ctx->input.orig_size;
             } else {
                 expected_required = artifact_round_up(ctx->file.size, MENDER_ARTIFACT_STREAM_BLOCK_SIZE) + MENDER_ARTIFACT_STREAM_BLOCK_SIZE;
--- a/core/src/mender-artifact.c
+++ b/core/src/mender-artifact.c
@@ -172,9 +172,14 @@ static mender_err_t process_artifact_data_callback(const char                   
  */
 static mender_err_t artifact_drop_file(mender_artifact_ctx_t *ctx);
 
-#ifdef CONFIG_MENDER_FULL_PARSE_ARTIFACT
-static mender_err_t artifact_read_header(mender_artifact_ctx_t *ctx);
-#endif /* CONFIG_MENDER_FULL_PARSE_ARTIFACT */
+/**
+ * @brief Shift data after parsing and update the respective checksum context
+ * @param checksum_key Key under which the checksum for the data is calculated/checked
+ * @param checksum_len Length of the data to include in the checksum update
+ * @return MENDER_OK if the function succeeds, error code otherwise
+ * @see artifact_shift_data()
+ */
+static mender_err_t artifact_shift_and_checksum_data(mender_artifact_ctx_t *ctx, size_t length, const char *checksum_key, size_t checksum_len);
 
 /**
  * @brief Shift data after parsing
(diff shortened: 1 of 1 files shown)


core/src/mender-artifact.c
```diff
@@ -1342,6 +1318,35 @@ artifact_drop_file(mender_artifact_ctx_t *ctx) {
...
+            /* Error already logged */
+            return MENDER_FAIL;
+        }
+
+        if (MENDER_OK != mender_sha256_update(checksum->context, ctx->input.data, checksum_len)) {
+            mender_log_error("Failed to update update checksum");
+            return MENDER_FAIL;
+        }
+    }
+#endif /* CONFIG_MENDER_FULL_PARSE_ARTIFACT */
```
larsewi, line 1345: Not sure if these would trigger unused argument warnings if compiled without `CONFIG_MENDER_FULL_PARSE_ARTIFACT`
```suggestion
    }
#else /* CONFIG_MENDER_FULL_PARSE_ARTIFACT */
    (void)checksum_key;
    (void)checksum_len;
#endif /* CONFIG_MENDER_FULL_PARSE_ARTIFACT */
```

larsewi (approved):
Awesome :rocket:

core/src/mender-artifact.c
```diff
@@ -358,9 +363,12 @@ mender_artifact_process_data(mender_artifact_ctx_t *ctx, void *input_data, size_
             new_size = ctx->input.length + input_length;
             /* data/ files are processed per block for which the original size of the buffer should
                be enough, but metadata is processed as whole files so there we expect we will need
-               more. */
+               more, except for header.tar (and tarballs in general) which are processed
+               transparently. */
             if (mender_utils_strbeginswith(ctx->file.name, "data/")) {
                 expected_required = ctx->input.orig_size;
+            } else if (mender_utils_strendswith(ctx->file.name, ".tar")) {
+                expected_required = ctx->input.orig_size;
```
lluiscampos, line 371: I think it would read cleaner: 

```suggestion
            if ((mender_utils_strbeginswith(ctx->file.name, "data/")) ||  (mender_utils_strendswith(ctx->file.name, ".tar"))) {
                expected_required = ctx->input.orig_size;
```

lluiscampos (approved):
Few nitpicks below, but it looks good :+1:

core/src/mender-artifact.c
```diff
@@ -172,9 +172,14 @@ static mender_err_t process_artifact_data_callback(const char
  */
 static mender_err_t artifact_drop_file(mender_artifact_ctx_t *ctx);
 
-#ifdef CONFIG_MENDER_FULL_PARSE_ARTIFACT
-static mender_err_t artifact_read_header(mender_artifact_ctx_t *ctx);
-#endif /* CONFIG_MENDER_FULL_PARSE_ARTIFACT */
+/**
+ * @brief Shift data after parsing and update the respective checksum context
+ * @param checksum_key Key under which the checksum for the data is calculated/checked
+ * @param checksum_len Length of the data to include in the checksum update
```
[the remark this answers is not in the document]
vpodzime: They were there, but then I replaced them with the `@see` line. I think that's good enough for a static function.

core/src/mender-artifact.c
```diff
@@ -539,6 +541,8 @@ artifact_parse_tar_header(mender_artifact_ctx_t *ctx) {
             return MENDER_OK;
         }
 
+        in_header_tar = (NULL != ctx->file.name) && StringEqual(ctx->file.name, "header.tar");
```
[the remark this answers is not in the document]
vpodzime: Yes, it would be something like `data/0000/header.tar`.

core/src/mender-artifact.c
vpodzime: Just by the look at the diffs I don't agree, but YMMV.

core/src/mender-artifact.c
vpodzime: I guess they would, good catch!

PR #127, opened 2025-01-08
Getting started documentation

danielskinstad:


README.md
```diff
@@ -105,50 +157,164 @@ After writing the code, you need to register the Update Module into the Mender M
...
+
+The Artifact Name is arbitrary, but should be unique.
+
+```bash
+UPDATE_MODULE=<update_module>
+ARTIFACT_NAME=<artifact_name>
+DEVICE_TYPE=<device_type>
+
+mender-artifact write module-image \
+  --type $ZEPHYR_IMAGE\
```
vpodzime, line 312: `$UPDATE_MODULE`, right?

vpodzime (commented):
Looks generally good to me. I didn't add comments for small things, leaving those for the final-ish review(s)

README.md
danielskinstad: oops, yup

README.md
```diff
@@ -28,32 +31,80 @@ The decision to fork the original mender-mcu-client was made to:
 * **Provide Official Support**: Ensure that the project receives the necessary attention and
   resources from Northern.tech to meet the needs of the community and enterprise users.
 
+## Getting started
```
lluiscampos, line 34: We should keep "Get started". We use to have "getting" in docs.mender.io and we switched it and we want to be consistent

README.md
```diff
@@ -28,32 +31,80 @@ The decision to fork the original mender-mcu-client was made to:
...
+In the `mender-mcu-integration` repository you will find:
 * A reference `west` workspace for Zephyr OS builds, including `CMake` and `KConfig` configurations.
 * Sample application for how to use `mender-mcu` APIs.
 * A list of boards that we have so far tested.
 
 We recommend starting there to understand how we have designed our solution and then coming back
 here for more in-depth information.
 
+### Compatibility
+| Zephyr OS version | Bootloader |
```
lluiscampos, line 50: I don't think the bootloader should be in this table. On the one hand, because we don't plan to integrate with any other bootloader, and in the other hand because it is an update module specific thing.

When introducing the update module (I haven't read so far, but let's assume there is a section for it) then we shall specify that "zephyr-image" update module the module installed by default that integrates with MCUboot

README.md
```diff
@@ -28,32 +31,80 @@ The decision to fork the original mender-mcu-client was made to:
...
+|-------------------|------------|
+| v3.7.0            | MCUboot    |
+
+### Boards
+The reference board for `mender-mcu` is the [ESP32-S3-DevKitC](https://docs.espressif.com/projects/esp-dev-kits/en/latest/esp32s3/esp32-s3-devkitc-1/index.html#).
```
lluiscampos, line 55: I would link to the Zephyr board page instead

README.md
```diff
@@ -28,32 +31,80 @@ The decision to fork the original mender-mcu-client was made to:
...
+
+To build the project, you need a board that supports MCUboot. The following link will filter
+the officially supported boards that also support MCUboot:
+
+* [Zephyr Project supported boards with MCU boot](https://docs.zephyrproject.org/latest/gsearch.html?q=MCUboot&check_keywords=yes&area=default#gsc.tab=0&gsc.q=MCUboot&gsc.ref=more%3Aboards&gsc.sort=)
+
```
lluiscampos, line 61: This is not true. Only if you want to use our module. Otherwise you can develop your own module using the provided API.

So this I would move to the section where we introduce Update Modules.

README.md
```diff
@@ -28,32 +31,80 @@ The decision to fork the original mender-mcu-client was made to:
...
+
+### Setting up a West project with `mender-mcu` as a module
```
lluiscampos, line 63: Here and in other places,

```suggestion
### Setting up a West project with `mender-mcu` as a Zephyr module
```

or

```suggestion
### Setting up a West project with `mender-mcu` as a west module
```

I would make sure to say "west module" to avoid confusion with "mender module" or "mender update module"

README.md
```diff
@@ -28,32 +31,80 @@ The decision to fork the original mender-mcu-client was made to:
...
+In order to use `mender-mcu` as a [module](https://docs.zephyrproject.org/latest/develop/modules.html#modules-external-projects)
+in a [West project](https://docs.zephyrproject.org/latest/develop/west/basics.html#west-workspace), you will
+need to add `mender-mcu` to your project's west manifest:
+
+```yaml
+manifest:
+  remotes:
+  #...
+    - name: mender
+      url-base: https://github.com/mendersoftware
```
lluiscampos, line 76: This is not necessary if we use `url` in the module. I think it is cleaner to remove the `remotes`, as we expect only one module to be needed, but also specially here in the README for a shorter snippet.

README.md
```diff
@@ -28,32 +31,80 @@ The decision to fork the original mender-mcu-client was made to:
...
+      revision: main
+      path: modules/mender-mcu
+      import: true
+```
+
+Then you will need to initialize a Zephyr workspace based on the manifest,
+which can be done like so:
+```
+west init workspace --manifest-url https://url/to/repository-containing-manifest
+cd workspace && west update
```
lluiscampos, line 90: As a user, assuming that I already have my hello-world project initialized, how can I add Mender MCU module and then just update?

This might be a better instruction than assuming we start from scratch. Or at least we should include both

README.md
```diff
@@ -28,32 +31,80 @@ The decision to fork the original mender-mcu-client was made to:
...
+and are prefixed with `MENDER_`.
+
+You can use either `west build -t menuconfig` or `west build -t guiconfig` to open
+the configuration menu.
+
+The following option **must** be set in any configuration:
+* `MENDER_SERVER_TENANT_TOKEN` which is a token that identifies which tenant a device belongs to.
+
+You change the server from the default `https://hosted.mender.io` by changing the following option:
+* `MENDER_SERVER_HOST`
```
lluiscampos, line 107: Instead of just the server, I would here say something like:

The most common options to modify are the following:
* MENDER_SERVER_HOST
* intervals
* ?

README.md
```diff
@@ -105,50 +157,173 @@ After writing the code, you need to register the Update Module into the Mender M
 update module function in [`mender-client.h`](include/mender-client.h).
```
lluiscampos, line 158: We are missing an explicit section on zephyr-image module (see comments below)

README.md
```diff
@@ -105,50 +157,173 @@ After writing the code, you need to register the Update Module into the Mender M
...
-### Dependencies
-- CMake
-- libcurl
-- cJSON
+### Certificates
+See this [example](https://github.com/mendersoftware/mender-mcu-integration/blob/main/src/utils/certs.c) from mender-mcu-integration for a demo.
 
-Example for Ubuntu/Debian:
+The `mender-mcu` client can take up to two certificates; if you connect to hosted Mender,
+you will need one for the server and for downloading Mender Artifacts.
```
lluiscampos, line 172:
```suggestion
The `mender-mcu` client can take up to two certificates; typically one for the Server API calls and for downloading Mender Artifacts from the Server storage. The latter is optional.
```

Something like that?

README.md
```diff
@@ -105,50 +157,173 @@ After writing the code, you need to register the Update Module into the Mender M
...
+
+The certificate should be in DER format.
+You will need to generate an INC file from the certificate.
+
```
lluiscampos, line 176: This is not necessarily true, you can the certs in different ways. I would be more generic and link to sample code of Zephyr project.

lluiscampos (commented):
(I run out of time, I'll continue later!)

lluiscampos (commented):
Done now with the review. Overall looks pretty comprehensive to me. You are covering everything that I listed in the JIRA task so from that side is good.

I added few more comments.

I suggest you first do now a round of small fixing (based on my comments) and then besides us the developers ping also our PO an docs owner for final review. We might get more feedback wrt the technical level of this guide and we might need to either write more stuff and/or create follow-up tasks to refactor some parts (example: I could see the PO suggesting to move the HM certs to mender-mcu)

Thank you for tackling this one :popcorn:

README.md
danielskinstad: I couldn't find any code samples from Zephyr - do you know of any?

README.md
```diff
@@ -74,81 +121,184 @@ always ready.
...
 inventory of the client and sent to the Mender Server in the next inventory report.
 
-See the source code for more.
-
+* User Provided Keys. The client can be configured to use a provided a keypair. If set to
+`NULL`, the client will auto-generate its own keypair using ECDSA.
 
 ### Update Modules API
 
+The Update Modules API is exposed in [mender-update-module.h](include/mender-update-module.h).
```
estenberg, line 129: Update Module API

README.md
```diff
@@ -74,81 +121,184 @@ always ready.
...
-```
-### Building the Client
+### Certificates
+See this [example](https://github.com/mendersoftware/mender-mcu-integration/blob/main/src/utils/certs.c) from mender-mcu-integration for a demo.
 
-1. Configure the build:
-```
-cmake -C cmake/CMake_posix_defaults.txt -B build tests/smoke
-```
+The `mender-mcu` client can take up to two certificates; typically one for the Server API calls, and
```
estenberg, line 170: Good you document this clearly but this seems like a limit we need to increase. If certificates need rotation we need to have the building blocks. I think we at least need support for 4 (two for each).

README.md
```diff
@@ -74,81 +121,184 @@ always ready.
...
+
+mender-artifact write module-image \
+  --type $UPDATE_MODULE \
+  --file build/zephyr/zephyr.signed.bin \
+  --artifact-name $ARTIFACT_NAME \
+  --device-type $DEVICE_TYPE \
+  --compression none
+```
 ### Deployment
 After creating and uploading the artifact to the server, you should be able to deploy it to the device.
```
estenberg, line 303: Would make this more confident :) "should be" doesn't instill confidence

... upload the Artifact to the Mender server and deploy it to your device.

README.md
```diff
@@ -28,32 +31,76 @@ The decision to fork the original mender-mcu-client was made to:
...
 
-## Get Started
-
-Since the project is under active development, we recommend watching the repository or checking back
-regularly for updates. Detailed documentation and usage instructions will be provided as the project
-progresses.
-
-
-## Get started with Zephyr OS
+For using `mender-mcu` for the first time, we strongly recommend you start with our [Zephyr reference application](https://github.com/mendersoftware/mender-mcu-integration).
```
estenberg, line 36: Not only that (recommend), but the rest of the doc assumes that mender-mcu-integration is being used. State this clearly and why (E.g. To simplify first-time testing, we will use ...)

estenberg (changes requested):
Added a few suggestions.

README.md
lluiscampos: I don't think this is true - or at least was not my intention.

The README here should enough to do your integration. There are examples (like the networking) that are taken from mender-mcu-integration, but it is not a hard requirement to use mender-mcu-integration to follow the instructions here.

Do you disagree @danielskinstad ?

danielskinstad: I do not disagree. Do you think my updated comment still stands based on your comment?

README.md
lluiscampos: This brings a point that we should rather move all the certs handling into mender-mcu core repo. Right now in the core we define if there will be one or two, but then is the application (mender-mcu-integration or any other user application) that will add the certs via `tls_credential_add(...)`. Instead, we can (and should?) have the mender client init function add the certs.

I suggest we create a task for this and do not modify the README further. In the task we shall write that the README needs to be updated accordingly.

README.md
lluiscampos: My point was that it is not necessary to be in DER format, if you configure via Kconfig to accept PEM format.

Example of installing certs with PEM format: https://github.com/zephyrproject-rtos/zephyr/tree/v3.7.1/samples/net/sockets/big_http_download/src

Now, DER format is more efficient, so we should not recommend PEM.

So, how could we do this generic enough here? Maybe: saying that the client expects the user to add the certificates using `tls_credential_add`. Docs at https://docs.zephyrproject.org/latest/doxygen/html/group__tls__credentials.html

(Also, in relation to [this other comment](https://github.com/mendersoftware/mender-mcu/pull/127#discussion_r1912909389), we probably will get the rid of this friction point anyway - if we move it into mender-mcu core, then we can require that the cert is already in DER format or support converting it from CMake ourselves)

README.md
vpodzime: I don't quite see how this is part of the `mender-mcu` (client) scope. In my eyes, this is part of the "environment", just like the network setup and other similar things.

vpodzime (approved):
Looks good to me.

README.md
lluiscampos: That is a valid point (and the original point for having it out of Mender hands).

But it gets tricky for us when down in the code `zsock_setsockopt` needs to know the CA tags (up to 4).

At least we should think about how to simplify this part :thinking: We could maybe have yet another callback to get this `TLS_SEC_TAG_LIST` from the user.

To move the topic forward: discuss in next client team meeting and create a task either for moving it to mender-mcu (most likely not) or other ideas for decoupling it a bit (like mine above or others).

README.md
lluiscampos: It's good.

lluiscampos (approved):

estenberg (approved):

PR #130, opened 2025-01-27
feat: Support platform-specific or custom allocators

vpodzime:
And use Zephyr-specific allocation functions on Zephyr by default, using either a separate heap (default) or the system heap.

This means we have to use our memory management functions instead of the standard ones everywhere, even if they can actually just call the standard ones.

Ticket: MEN-7808
Changelog: none

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -20,7 +20,7 @@ test:static:license:
   script:
     - ./.github/workflows/check_headers.sh
 
-test:static:format:
+test:static:code:
   extends: .test:static
   script:
     - ./.github/workflows/check_include_guards.sh


vpodzime:
Ready for a review. Now that it seems to finally work, I need to experiment with the minimum heap size we can use and adapt the default values in the *Kconfig*.

cmake/mender_mcu_sources.txt
```diff
@@ -95,6 +95,12 @@ if (NOT CONFIG_MENDER_PLATFORM_NET_TYPE)
 else()
     message(STATUS "Using custom '${CONFIG_MENDER_PLATFORM_NET_TYPE}' platform network implementation")
 endif()
+if (NOT CONFIG_MENDER_PLATFORM_MEM_TYPE)
+    message(STATUS "Using default 'generic' platform memory allocation implementation")
+    set(CONFIG_MENDER_PLATFORM_MEM_TYPE "generic")
+else()
+    message(STATUS "Using custom '${CONFIG_MENDER_PLATFORM_NET_TYPE}' platform network implementation")
```
lluiscampos, line 102:
```suggestion
    message(STATUS "Using custom '${CONFIG_MENDER_PLATFORM_MEM_TYPE}' platform memory allocation implementation")
```

Something like that

include/mender-alloc.h
```diff
@@ -0,0 +1,48 @@
...
+typedef void *(*MenderReallocator)(void *ptr, size_t size);
+typedef void (*MenderDeallocator)(void *ptr);
+
+void mender_set_allocation_funcs(MenderAllocator mender_malloc_func, MenderReallocator mender_realloc_func, MenderDeallocator free_func);
+void mender_set_platform_allocation_funcs(void);
+
+void *mender_malloc(size_t size);
+void *mender_calloc(size_t n, size_t size);
+void *mender_realloc(void *ptr, size_t size);
+void  mender_free(void *ptr);
```
lluiscampos, line 42: Nitpick,

These could have simple doc just saying that they follow the same prototype that the libc malloc/calloc/etc

core/src/mender-alloc.c
```diff
@@ -0,0 +1,68 @@
...
+
+static MenderAllocator   malloc_fn  = NULL;
+static MenderReallocator realloc_fn = NULL;
+static MenderDeallocator free_fn    = NULL;
+
+void
+mender_set_allocation_funcs(MenderAllocator malloc_func, MenderReallocator realloc_func, MenderDeallocator free_func) {
+    malloc_fn  = malloc_func;
+    realloc_fn = realloc_func;
+    free_fn    = free_func;
```
lluiscampos, line 33: Should we allow passing only one of the three?

As it is now because we just check individually if they are NULL when using them (ie mender_malloc checks or calls malloc) we could end-up with a user specifying only one and having really bad memory allocation (like custom malloc and libc realloc, or custom malloc and no custom free :lady_beetle: )

I suggest we assert on this - it should be caught by a developer at dev time

**Edit:** I see now that this is handled in mender-client code. However, with the future vision of using the project also as an SDK I think we should move the checks here.

core/src/mender-utils.c
```diff
@@ -119,6 +119,70 @@ mender_utils_strrstr(const char *haystack, const char *needle) {
...
+    *result = mender_malloc((size_t)len + 1);
+    if (NULL == *result) {
+        return -1;
+    }
+    va_copy(ap_copy, ap);
+    len = vsnprintf(*result, len + 1, fmt, ap_copy);
+    va_end(ap_copy);
+
+    return len;
+}
```
lluiscampos, line 184: I am not good in `va` magic - but it reads good to me :+1:

include/mender-alloc.h
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Standard parameters and semantics apply to the three types and functions below.
+ */
+typedef void *(*MenderAllocator)(size_t size);
+typedef void *(*MenderReallocator)(void *ptr, size_t size);
+typedef void (*MenderDeallocator)(void *ptr);
+
+void mender_set_allocation_funcs(MenderAllocator mender_malloc_func, MenderReallocator mender_realloc_func, MenderDeallocator free_func);
+void mender_set_platform_allocation_funcs(void);
```
lluiscampos, line 37: These two need proper doc :pray:

platform/memory/zephyr/src/mender-alloc.c
```diff
@@ -0,0 +1,58 @@
...
+
+static void
+mender_zephyr_free(void *ptr) {
+#ifdef CONFIG_MENDER_SEPARATE_HEAP
+    k_heap_free(&mender_heap, ptr);
+#else
+    k_free(ptr);
+#endif
+}
+
```
lluiscampos, line 54: Very good :clap:

target/zephyr/Kconfig
```diff
@@ -140,6 +140,23 @@ if MENDER_MCU_CLIENT
...
+            config MENDER_PLATFORM_MEM_TYPE_DEFAULT
+                bool "default"
+            config MENDER_PLATFORM_MEM_TYPE_GENERIC
+                bool "generic"
+        endchoice
+
+        config MENDER_PLATFORM_MEM_TYPE
+            string
+            default "zephyr" if MENDER_PLATFORM_LOG_TYPE_DEFAULT
+            default "generic" if MENDER_PLATFORM_LOG_TYPE_GENERIC
```
lluiscampos, line 158:
```suggestion
            default "zephyr" if MENDER_PLATFORM_MEM_TYPE_DEFAULT
            default "generic" if MENDER_PLATFORM_MEM_TYPE_GENERIC
```

target/zephyr/Kconfig
```diff
@@ -230,6 +247,36 @@ if MENDER_MCU_CLIENT
...
+
+            config MENDER_HEAP_SIZE
+               int "Mender heap size (kB)"
+               depends on MENDER_SEPARATE_HEAP
+               default 10
+               range 4 32
+               help
+                   Size of the Mender heap.
+
+            config HEAP_MEM_POOL_ADD_SIZE_MENDER
```
lluiscampos, line 268: Unused.

lluiscampos (commented):
This looks very good! Nicely solved.

See my comments below. Besides small types, nothing really blocking that I could find.

larsewi (commented):
Looks good so far :rocket: Would be nice to add a GitHub Action to check that `malloc()`, `realloc()` & `free()` are not used anywhere other than _mender-alloc.c_ ofc

target/zephyr/Kconfig
vpodzime: This is a trick, as explained on Slack and when demoing. I added a comment to make it clearer.

include/mender-alloc.h
vpodzime: It's there, right above the function type definitions.

core/src/mender-alloc.c
vpodzime: > However, with the future vision of using the project also as an SDK I think we should move the checks here.

In that future version, I don't want this function to be part of the SDK/public API. The only way to set the allocation functions should be via `mender_client_init()`. Otherwise we ask for trouble because calling this function at some other point would cause issues too.

vpodzime:
> See my comments below. Besides small types, nothing really blocking that I could find.

Comments addressed. The only remaining thing here that I plan to do is the GH actions check suggested by Lars.

core/src/mender-alloc.c
lluiscampos: Okey, makes sense. Thank you for the reasoning.

include/mender-alloc.h
```diff
@@ -0,0 +1,57 @@
...
+#define __MENDER_ALLOC_H__
+
+#ifdef __cplusplus
+extern "C" {
+#endif /* __cplusplus */
+
+#include <stddef.h> /* size_t */
+
+/**
+ * Standard parameters and semantics apply to the three types and functions below.
```
lluiscampos, line 30:
```suggestion
 * Standard parameters and semantics apply to the types and functions below.
```

:thinking: The "three" there is what confused me in the first round. I am not a grammar expert but "three apples and pears translates in my head to "three apples and three pears"

lluiscampos (approved):
Thank you for adding the check. Well done!

(I am working on fixing CI independently of this PR)

include/mender-alloc.h
vpodzime: Easy fix then! :)

include/mender-alloc.h
```diff
@@ -0,0 +1,57 @@
...
+void *mender_calloc(size_t n, size_t size);
+void *mender_realloc(void *ptr, size_t size);
+void  mender_free(void *ptr);
+
+/**
+ * @brief Set memory allocation functions for Mender to use
+ * @warning This function has to be called before Mender allocates any memory
+ * @note  All three parameters have to be non-%NULL
+ */
+void mender_set_allocation_funcs(MenderAllocator mender_malloc_func, MenderReallocator mender_realloc_func, MenderDeallocator free_func);
```
larsewi, line 46: One of the parameters does not have the `mender_` suffix. Does not really matter if it is there or not, but consistency would be nice

include/mender-client.h
```diff
@@ -54,6 +55,11 @@ typedef struct {
     uint32_t inventory_update_interval; /**< Inventory update interval, default is compile-time defined */
 #endif                                  /* CONFIG_MENDER_CLIENT_INVENTORY */
     bool recommissioning;               /**< Used to force creation of new authentication keys */
+    struct {
+        MenderAllocator   mender_malloc_func;
+        MenderReallocator mender_realloc_func;
+        MenderDeallocator free_func;
```
larsewi, line 61: Same here (i.e., no `mender_` suffix). Maybe this was the intention?

larsewi (approved):

include/mender-client.h
vpodzime: No it wasn't. It was me using `sed s/free/mender_free/` multiple times, accidentally, and then trying to fix it. Good :eyes: !

danielskinstad (approved):
Nice :+1:

PR #131, opened 2025-01-30
fix: check that the update module is set in mender_download_artifact

danielskinstad:
Due to `artifact_read_data_prepare` not being called when an artifact doesn't have any files, the update module might be NULL after `mender_http_artifact_download` returns MENDER_OK.

Add an explicit check to ensure this is caught.

Changelog: None
Ticket: MEN-7804


core/src/mender-artifact-download.c
```diff
@@ -53,6 +53,12 @@ mender_download_artifact(const char *uri, mender_deployment_data_t *deployment_d
         return ret;
     }
 
+    /* If the artifact doesn't contain a file, the update module won't be set */
```
lluiscampos, line 56: Discussed face to face,

An Artifact with no files is a valid use case of an "update" and the MCU Client should handle that.

Without looking at the code, my gut feeling is that setting we are setting `dl_data.update_module` too late (when parsing files) while we should do it as soon as we knot the artifact type

lluiscampos (changes requested):

danielskinstad:
Didn't find a better/simpler way to check if the artifact had a file - lmk if you have a better solution

core/src/mender-artifact.c
```diff
@@ -757,6 +764,9 @@ artifact_read_manifest(mender_artifact_ctx_t *ctx) {
         const char *checksum_str = line;
         const char *filename     = separator + 2;
 
+        /* Store the filename in the artifact */
+        artifact_file = (char *)filename;
```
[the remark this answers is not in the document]
vpodzime: It also doesn't create a copy so it relies on pure luck of the data staying in place.

core/src/mender-zephyr-image-update-module.c
```diff
@@ -136,6 +137,14 @@ mender_zephyr_image_set_pending_image(MENDER_NDEBUG_UNUSED mender_update_state_t
     assert(MENDER_UPDATE_STATE_INSTALL == state);
     mender_err_t ret;
 
+#ifdef CONFIG_MENDER_FULL_PARSE_ARTIFACT
+    const char *artifact_filename = mender_artifact_get_filename();
```
vpodzime, line 141: Why store the name when we only need to know if there was a payload file or not?

core/src/mender-artifact.c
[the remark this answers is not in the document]
danielskinstad: Yeah - implemented this in the last second after I moved the update module setting. This will, as you say, be overwritten - and it won't be the file we want. I saw filename in the function and immediatly thought of the payload file 😅

core/src/mender-zephyr-image-update-module.c
danielskinstad: Yup, a boolean or pehaps a counter seems more reasonable

vpodzime: I thought about this a bit more and I think this should be handled in the update module itself. It can "remember" itself if it wrote/flashed a file or not and behave accordingly.

core/src/mender-zephyr-image-update-module.c
```diff
@@ -136,6 +139,12 @@ mender_zephyr_image_set_pending_image(MENDER_NDEBUG_UNUSED mender_update_state_t
     assert(MENDER_UPDATE_STATE_INSTALL == state);
     mender_err_t ret;
 
+    if (!artifact_has_payload) {
+        mender_log_error("No file in artifact");
```
vpodzime, line 143: *No payload in artifact*, maybe?

core/src/mender-zephyr-image-update-module.c
```diff
@@ -29,6 +29,8 @@
  */
 static void *mcu_boot_flash_handle = NULL;
 
+static bool artifact_has_payload;
```
vpodzime, line 32: I'd actually name this `artifact_had_payload` as that reflects that it's only valid after an artifact was processed.

vpodzime (approved):
Looks good to me either way.

core/src/mender-zephyr-image-update-module.c
danielskinstad: Sure

core/src/mender-zephyr-image-update-module.c
```diff
@@ -126,6 +128,7 @@ mender_zephyr_image_download_artifact_flash_callback(MENDER_NDEBUG_UNUSED mender
         }
     }
 
+    artifact_had_payload = true;
```
lluiscampos, line 131: Shouldn't this be done only if `dl_data->offset + dl_data->length >= dl_data->size`? iow, here we might signal that the payload has been flashed even on short writes.

core/src/mender-zephyr-image-update-module.c
```diff
@@ -29,6 +29,8 @@
  */
 static void *mcu_boot_flash_handle = NULL;
 
+static bool artifact_had_payload;
```
lluiscampos, line 32: (in the context of my previous comment)

The variable name reads strange to me,

what about `mcu_boot_image_flashed` or something like that? The prefix `mcu_boot`for consistency with the above and the name `image_flashed` to relate it to the action (flashing) rather than the trigger or source (the payload)

lluiscampos (approved):
It is technically correct, but I think we can make it more strict. Looks good either way :rocket: 

And sorry that I have been blocking this one a couple of days :see_no_evil:

core/src/mender-zephyr-image-update-module.c
vpodzime: But this is not to capture the fact that the payload was flashed, right? It's only to capture the fact that there was some payload in the artifact. Should the update module care about whether flashing was successful or not? It should, of course, report an error if flashing fails. But then it's the state machine's responsibility to call the proper callbacks, isn't it?

lluiscampos: > But this is not to capture the fact that the payload was flashed, right? 

That is what I understand that we are doing in `mender_zephyr_image_set_pending_image` -> to fail if there was "no payload" but really signaling that there was no flash.

> It's only to capture the fact that there was some payload in the artifact. Should the update module care about whether flashing was successful or not?

This doesn't make sense to me.

@danielskinstad what are we trying to fix here in the first place?

>  It should, of course, report an error if flashing fails. But then it's the state machine's responsibility to call the proper callbacks, isn't it?

Yes, exactly. But then I don't know what this code is about.

danielskinstad: The task said that the error message was not explicit enough; so the "fix" in the update-module is to try to explicitly log what's happening.

lluiscampos: Ok. Let's merge as-is :+1:

PR #133, opened 2025-01-30
ci: Prevent Debian/Ubuntu from interactively installing packages

lluiscampos:
A recent update of `tzdata`, which requires user input to set timezone, triggered this problem in CI shown as "broken pipe".

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -8,6 +8,10 @@ stages:
   - test
   - publish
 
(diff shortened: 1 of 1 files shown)


vpodzime (approved):
ACK

lluiscampos:
A temporary fix has been submitted in #136, while the other relevant changes of this PR has been submitted into #137 :+1:

PR #134, opened 2025-01-30
feat: Fall back to unconditional reboot if reboot callback fails

vpodzime:
Plus one more case that we want to handle the same way.


larsewi (approved):

danielskinstad (approved):
Looks good to me

PR #139, opened 2025-01-31
chore: Rename mender-scheduler to mender-os and adapt function names

vpodzime:
Things like `mender_scheduler_mutex_create()` and
`mender_scheduler_reboot()` look/sound quite confusing even though, strictly speaking, both mutual exclusion and reboot are related to scheduling.

Ticket: none
Changelog: none

### DIFF
--- a/include/mender-os.h
+++ b/include/mender-os.h
@@ -1,6 +1,6 @@
 /**
- * @file      mender-scheduler.h
- * @brief     Mender scheduler interface
+ * @file      mender-os.h
+ * @brief     Mender OS interface
  *
  * Copyright joelguittet and mender-mcu-client contributors
  * Copyright Northern.tech AS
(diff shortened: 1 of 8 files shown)


danielskinstad (approved):
Looks good to me

include/mender-os.h
```diff
@@ -1,5 +1,5 @@
 /**
- * @file      mender-scheduler.h
+ * @file      mender-os.h
  * @brief     Mender scheduler interface
```
larsewi, line 3: Should it be?
```suggestion
 * @brief     Mender OS interface
```
If so, also change in platform specific header files

larsewi (approved):

include/mender-os.h
vpodzime: Another good catch by Larsito! :beers:

larsewi: :beers:

PR #140, opened 2025-01-31
ci: trigger mender-mcu-pipeline

danielskinstad:


.gitlab-ci.yml
```diff
@@ -4,8 +4,12 @@ include:
       - '.gitlab-ci-check-commits.yml'
       - '.gitlab-ci-github-status-updates.yml'
 
+variables:
+  MENDER_MCU_REVISION: main
```
vpodzime, line 8: We need the PR revision here, right?

vpodzime (commented):
Very nice! Let's get this working ASAP, please. :)

.gitlab-ci.yml
danielskinstad: Yup - that should be the only thing missing here

vpodzime (approved):
Can't say I fully understand the YAML magic, but looks good to me.

.gitlab-ci.yml
```diff
@@ -127,6 +131,16 @@ test:static:release:
   variables:
     BUILD_TYPE: Release
 
+trigger:mender-mcu-integration:
+  rules:
+  - if: $CI_COMMIT_BRANCH =~ /^(pr_[0-9]+)/
+    variables:
+      MENDER_MCU_REVISION: ${CI_EXTERNAL_PULL_REQUEST_SOURCE_BRANCH_SHA}
```
lluiscampos, line 138: TIL :tada: Thank you

.gitlab-ci.yml
```diff
@@ -127,6 +131,16 @@ test:static:release:
   variables:
     BUILD_TYPE: Release
 
+trigger:mender-mcu-integration:
+  rules:
+  - if: $CI_COMMIT_BRANCH =~ /^(pr_[0-9]+)/
```
lluiscampos, line 136:
```suggestion
  - if: $CI_COMMIT_BRANCH =~ /^(pr_[0-9]+)$/
```

:point_up: if we want to be even more strict

.gitlab-ci.yml
```diff
@@ -127,6 +131,16 @@ test:static:release:
...
+    variables:
+      MENDER_MCU_REVISION: ${CI_EXTERNAL_PULL_REQUEST_SOURCE_BRANCH_SHA}
+  - when: always
+  trigger:
```
lluiscampos, line 140: I would add to this trigger `branch` with a variable, being `main` by default, so that we can at least manually trigger a build with a PR in both repos (example: a change in mender-mcu requires changes in mender-mcu-integration - without the ability of running a pipeline with mender-mcu-integration versioned I will never get a green pipeline in mender-mcu)

lluiscampos (commented):
See suggestion below wrt mender-mcu-integration versioning

.gitlab-ci.yml
```diff
@@ -131,6 +136,17 @@ test:static:release:
   variables:
     BUILD_TYPE: Release
 
+trigger:mender-mcu-integration:
+  rules:
+  - if: $CI_COMMIT_BRANCH =~ /^(pr_[0-9]+)$/
```
lluiscampos:
```suggestion
  - if: $CI_COMMIT_BRANCH =~ /^pr_[0-9]+$/
```

:point_up: we don't need the grouping

lluiscampos (approved):

.gitlab-ci.yml
danielskinstad: Am I doing something wrong with the branch? It can't find the mender-mcu-integration reference now

.gitlab-ci.yml
danielskinstad: Nice catch

PR #143, opened 2025-02-06
fix: Allow `null` values in deployment data

vpodzime:
Some of the fields in deployment data are initialized to `null` because they are populated later from other places. Not allowing `null` values in such fields thus means that freshly created and initialized deployment data is not valid until those fields are populated. The data is validated every time we try to store it in non-volatile memory so it either needs to never be saved until being fully populated or we need to consider not-fully-populated deployment data valid. However, since the state of the client is saved as part of the deployment data, the latter approach is required because otherwise we can't even save the client's state in some cases.

Ticket: MEN-7934
Changelog: none

### DIFF
--- a/core/src/mender-deployment-data.c
+++ b/core/src/mender-deployment-data.c
@@ -40,6 +40,11 @@
  */
 #define MAX_STATE_DATA_STORE_COUNT 28.0
 
+static cJSON_bool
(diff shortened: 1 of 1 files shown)


lluiscampos (approved):
I run into this in the past, long time ago. It makes sense to me (after reading a great reasoning in the commit it is difficult not to :laughing: )

PR #144, opened 2025-02-07
Fix: Fail deployment in case of checksum mismatch

vpodzime:
The first commit is from #143, only the last two commits are relevant for this PR.


lluiscampos (approved):
It looks correct.

danielskinstad (approved):
Looks good to me

PR #148, opened 2025-02-13
ci: use `CI_COMMIT_SHA` as default value for `MENDER_MCU_REVISION`

danielskinstad:
If we manually trigger a pipeline on a PR branch,
`CI_EXTERNAL_PULL_REQUEST_SOURCE_BRANCH_SHA` will be empty, resulting in the downstream pipeline to fail.

Furthermore, specifying which revision of mender-mcu to use, is done by selecting which branch to run the pipeline for. Since we need the sha to be able to checkout PR branches in mender-mcu-integration, we can simply always set `MENDER_MCU_REVISION` to the sha of the latest commit.

Add a description to the variables you can modify to make it easier when manually creating pipelines.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -11,9 +11,12 @@ stages:
 
 variables:
   # mender-artifact version for tests
-  MENDER_ARTIFACT_VERSION: 4.0.0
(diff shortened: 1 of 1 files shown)


lluiscampos (approved):
Good call.

PR #149, opened 2025-02-14
MEN-7692: Split headers and functions between public and private ones

vpodzime:


vpodzime:
I need to adapt `check_include_guards.sh` again. Private headers use the `_PRIV` suffix for the guards to avoid conflicts.

vpodzime:
>  ci/gitlab — Pipeline failed on Gitlab 

There's only one failure, caused by https://github.com/mendersoftware/mender-mcu-integration/pull/48 required to be included in the build.

src/include/artifact-download-data.h
```diff
@@ -17,16 +17,16 @@
  * limitations under the License.
  */
 
-#ifndef __MENDER_ARTIFACT_DOWNLOAD_DATA_H__
-#define __MENDER_ARTIFACT_DOWNLOAD_DATA_H__
+#ifndef __MENDER_ARTIFACT_DOWNLOAD_DATA_PRIV_H__
+#define __MENDER_ARTIFACT_DOWNLOAD_DATA_PRIV_H__
```
lluiscampos, line 21: It is a good call to add the `_PRIV_` to avoid collisions.

One small thing slightly confusing is that for the headers that have both public and private versions (example, `alloc.h`) we have the explicit `(private API)` in the doc line and the  `_PRIV_` in the guard, while for the headers that are only private, like the one I am commenting on, we only have the guard modified.

If it's just one more ninja command to run, could we add the `(private API)` in the doc line also in these?

lluiscampos (approved):
Reviewing commit-by-commit was a breeze. Than kyou :rocket:

It looks very good to me. I made a comment for your consideration.

src/include/artifact-download-data.h
vpodzime: Good idea. I actually didn't even have the `_PRIV_` in the guard in those private-only headers. But that complicated the `check_include_headers.h` check too much. So I decided to add it to all private headers because it cannot hurt. :) Adding the `(private API)` part is easy and it can potentially be helpful to people looking at the sources.

PR #152, opened 2025-02-18
MEN-8057: Memory management fixes (round 1)

vpodzime:
These are the fixes for memory management issues (mostly leaks) that happen during new deployment checking and inventory updating. Next step (a follow-up PR) is to check and fix the ones happening during deployment download and installation.

### DIFF
--- a/src/core/api.c
+++ b/src/core/api.c
@@ -405,6 +405,7 @@ END:
     mender_utils_key_value_list_free(provides);
 #endif /* CONFIG_MENDER_FULL_PARSE_ARTIFACT */
 #endif /* CONFIG_MENDER_PROVIDES_DEPENDS */
+    mender_free((void *)artifact_name);
     cJSON_Delete(json_payload);
     mender_free(payload);
     return ret;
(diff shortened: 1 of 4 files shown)


danielskinstad (approved):
Nice

src/core/api.c
```diff
@@ -706,6 +706,7 @@ mender_api_publish_inventory_data(mender_keystore_t *inventory) {
 END:
 
     /* Release memory */
+    mender_free((void *)artifact_name);
```
larsewi, line 709: Should not `artifact_name` be defined as a non-const pointer if it is allocated on the heap?

vpodzime: Why? `const` mean the data should not be changed.

lluiscampos (approved):
These fixes look good to me

src/core/api.c
larsewi: When I see `const` pointer, my initial thought is that what ever this points to, I'm not the owner of it, and I should not worry about freeing it.

When ever I see casting of a `const` pointer to a non-`const` pointer, alarm bells go off in my head.

However, my thinking may be wrong. So either way is fine by me.

PR #154, opened 2025-02-25
fix: allocate `artifact_name` if it's unknown

danielskinstad:


src/platform/storage/zephyr/nvs/storage.c
```diff
@@ -383,7 +383,7 @@ mender_storage_get_artifact_name(const char **artifact_name) {
     mender_err_t ret = nvs_read_alloc(&mender_storage_nvs_handle, MENDER_STORAGE_NVS_ARTICACT_NAME, (void **)artifact_name, &artifact_name_length);
     if (MENDER_OK != ret) {
         if (MENDER_NOT_FOUND == ret) {
-            *artifact_name = "unknown";
+            *artifact_name = mender_utils_strdup("unknown");
```
lluiscampos, line 386: I think it is better that @vpodzime gives the final :+1: here. The fix looks correct to me, but I am very confident that I tested this long time ago - so it must be related to other PRs optimizing things which I lost the overview of.

danielskinstad: Vratislav fixed some memory leaks last week, and amongst them: actually freeing artifact name. I guess that's why it hasn't been caught before.

lluiscampos (approved):

src/platform/storage/zephyr/nvs/storage.c
vpodzime: Exactly. I recall noticing this as a potential issue back then when it was introduced, but we were not `free()`ing `artifact_name` anywhere at that point. Anyway, this is the right fix.

PR #157, opened 2025-02-27
MEN-8075: Do early integrity checks for the non-payload items

lluiscampos:
We shall check `version` and `header.tar` as soon as they are fully downloaded, so that the rest of the download can be aborted if the integrity of any of these is compromised.

Implemented by splitting the integrity checks into individual items (to be used during header parsing) and the remaining ones (to be used after the payload(s) have been downloaded).

### DIFF
--- a/src/core/artifact.c
+++ b/src/core/artifact.c
@@ -972,7 +972,7 @@ artifact_read_type_info(mender_artifact_ctx_t *ctx) {
         ret = MENDER_FAIL;
         goto END;
     }
-#ifdef CONFIG_MENDER_FULL_PARSE_ARTIFACT
+
     cJSON *json_provides = cJSON_GetObjectItemCaseSensitive(object, "artifact_provides");
     if (cJSON_IsObject(json_provides)) {
         if (MENDER_FAIL == artifact_parse_provides_depends(json_provides, &(ctx->payloads.values[index].provides))) {
@@ -1021,7 +1021,6 @@ artifact_read_type_info(mender_artifact_ctx_t *ctx) {
             }
         }
     }
-#endif
 
     /* Shift data in the buffer */
     /* header.tar has a checksum entry in the manifest as a whole */
--- a/src/core/artifact.c
+++ b/src/core/artifact.c
@@ -254,46 +254,67 @@ artifact_checksum_get_or_create(mender_artifact_ctx_t *ctx, const char *filename
 }
 #endif /* CONFIG_MENDER_FULL_PARSE_ARTIFACT */
 
-mender_err_t
-mender_artifact_check_integrity(mender_artifact_ctx_t *ctx) {
-    assert(NULL != ctx);
-
+static mender_err_t
+is_checksum_valid(mender_artifact_checksum_t *checksum) {
 #ifdef CONFIG_MENDER_FULL_PARSE_ARTIFACT
-    for (mender_artifact_checksum_t *checksum = ctx->artifact_info.checksums; NULL != checksum; checksum = checksum->next) {
-        unsigned char computed[MENDER_DIGEST_BUFFER_SIZE];
-        mender_log_debug("Checking integrity for artifact file '%s'", checksum->filename);
+    unsigned char computed[MENDER_DIGEST_BUFFER_SIZE];
+    mender_log_debug("Checking integrity for artifact file '%s'", checksum->filename);
 
-        if (MENDER_OK != mender_sha256_finish(checksum->context, computed)) {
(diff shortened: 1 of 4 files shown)


lluiscampos:
I want to expand a bit the unit tests and also do more manual testing, but the main design is up there so please review :smiley:

src/core/artifact.c
```diff
@@ -254,49 +254,74 @@ artifact_checksum_get_or_create(mender_artifact_ctx_t *ctx, const char *filename
...
+mender_artifact_check_integrity_item(mender_artifact_ctx_t *ctx, const char *filename) {
+    assert(NULL != ctx);
+    bool found = false;
+    for (mender_artifact_checksum_t *checksum = ctx->artifact_info.checksums; NULL != checksum && !found; checksum = checksum->next) {
+        if (StringEqual(filename, checksum->filename)) {
+            found             = true;
+            checksum->checked = true;
+            if (MENDER_OK != is_checksum_valid(checksum)) {
+                return MENDER_FAIL;
+            }
```
vpodzime, line 306: Why not `return is_checksum_valid(checksum);` instead of the whole `if`? With that you could also do `return MENDER_FAIL` below because it would only be reached when not found. There cannot be multiple checksums for the same file name, right?

src/core/artifact.c
```diff
@@ -420,6 +445,12 @@ mender_artifact_process_data(mender_artifact_ctx_t *ctx, void *input_data, size_
 
                 /* Read manifest file */
                 ret = artifact_read_manifest(ctx);
+
+                /* Early integrity check for version file */
+                if (MENDER_OK != mender_artifact_check_integrity_item(ctx, "version")) {
```
vpodzime, line 450: I'd add `(ret == MENDER_OK) &&` to the condition because we don't want to do integrity check if the above function fails.

src/include/artifact.h
```diff
@@ -131,12 +132,20 @@ mender_err_t mender_artifact_get_ctx(mender_artifact_ctx_t **ctx);
 mender_err_t mender_artifact_process_data(mender_artifact_ctx_t *ctx, void *input_data, size_t input_length, mender_artifact_download_data_t *dl_data);
 
 /**
- * @brief Do integrity checks by comparing the manifest checksums to the computed ones
+ * @brief Do integrit check to one specific item by comparing the manifest checksum to the computed one.
```
vpodzime, line 135: This needs to be more gritty :wink:

src/include/artifact.h
```diff
@@ -63,6 +63,7 @@ typedef struct mender_artifact_checksum_t mender_artifact_checksum_t;
 struct mender_artifact_checksum_t {
     char                       *filename;
     unsigned char               manifest[MENDER_DIGEST_BUFFER_SIZE];
+    bool                        checked;
```
vpodzime, line 66: An alternative is to actually drop the checked checksums from the list. I'd prefer that because we don't need them after being checked and it's quite some heap space to get back.

vpodzime (commented):
Looks good to me otherwise.

lluiscampos (commented):
Thank you both for your good reviews. I will apply all of the suggestions (together with adding tests) tomorrow :+1:

src/include/artifact.h
lluiscampos: @vpodzime I did this on a separate commit to ease your review.

larsewi (approved):
Simpático 🚀

tests/unit/core/artifact_test.cpp
```diff
@@ -234,29 +234,88 @@ TEST_F(MenderArtifactTest, ProcessData_EmptyDownloadData) {
...
+}
+
+TEST_F(MenderArtifactTest, CheckIntegrityBadChecksums) {
+
+    /* Artifacts created with commands like the following:
+    mender-artifact write module-image --type unit-test --file LICENSE --artifact-name test-bad-version --device-type unit-test --compression none
+    tar -xf artifact.mender
+    sed -i '/version/ s/^[0-9a-f]\{8\}/deadbeef/' manifest 
+    tar -cf test-bad-version.artifact version manifest header.tar data/0000.tar 
+    */
```
danielskinstad, line 282: Just a suggestion, but instead of commiting the artifacts to the repo, I think it would potentially easy maintenance in the future if we just generate and modify the artifacts in the test. You should be able to just call `CreateArtifact` with those exact commands, as the function can take a custom script and execute it.

danielskinstad (approved):
Looks good to me

PR #158, opened 2025-02-28
chore: always go cleanup->end in download state

danielskinstad:
This is more of a suggestion and something that comes from me working on state machine tests.
I have relied on using `MENDER_UPDATE_STATE_END` as a way to know when when the state transitions are over. Unifying the flow to always go (some flow)->`MENDER_UPDATE_STATE_CLEANUP`->`MENDER_UPDATE_STATE_END`, instead of having exceptions for certain states, makes it a bit easier.

Let me know if it makes sense.

### DIFF
--- a/src/core/client.c
+++ b/src/core/client.c
@@ -1047,11 +1047,8 @@ mender_client_update_work_function(void) {
                     }
                 } else {
                     mender_log_error("Unable to download artifact");
-                    if (NULL == mender_update_module) {
-                        /* Error logged in mender_client_download_artifact_callback() */
-                        mender_client_publish_deployment_status(deployment->id, MENDER_DEPLOYMENT_STATUS_FAILURE);
-                        goto END;
-                    }
+                    /* Error logged in mender_client_download_artifact_callback() */
+                    ret = MENDER_FAIL;
                 }
                 if (MENDER_OK != ret) {
                     mender_client_publish_deployment_status(deployment_id, MENDER_DEPLOYMENT_STATUS_FAILURE);
@@ -1140,8 +1137,12 @@ mender_client_update_work_function(void) {
                 /* fallthrough */
 
             case MENDER_UPDATE_STATE_CLEANUP:
-                if (NULL != mender_update_module->callbacks[update_state]) {
-                    ret = mender_update_module->callbacks[update_state](update_state, (mender_update_state_data_t)NULL);
+                if (NULL != mender_update_module) {
+                    if (NULL != mender_update_module->callbacks[update_state]) {
+                        ret = mender_update_module->callbacks[update_state](update_state, (mender_update_state_data_t)NULL);
+                    }
+                } else {
+                    ret = MENDER_FAIL;
                 }
                 NEXT_STATE;
                 mender_storage_delete_deployment_data();


src/core/client.c
```diff
@@ -1044,11 +1044,8 @@ mender_client_update_work_function(void) {
                     }
                 } else {
                     mender_log_error("Unable to download artifact");
-                    if (NULL == mender_update_module) {
-                        /* Error logged in mender_client_download_artifact_callback() */
-                        mender_client_publish_deployment_status(deployment->id, MENDER_DEPLOYMENT_STATUS_FAILURE);
-                        goto END;
-                    }
+                    /* Error logged in mender_client_download_artifact_callback() */
+                    ret = MENDER_FAIL;
```
vpodzime, line 1048: This relies on the fact that the failure transition is to the CLEANUP state that handles no update module gracefully. But if we introduce some intermediate state or whatever, it will break. The `goto END` was/is there to make sure no state transitions happen and thus no UM is needed.

danielskinstad: > if we introduce some intermediate state or whatever, it will break.

Right, that's true. 

What about going directly to `MENDER_UPDATE_STATE_END` and skipping CLEANUP? I can probably find a way in the tests without altering this, but comparing random lines in stdout and requiring 10 extra lines of code for one specific state transition would be nice to avoid, but I won't push any further on this

vpodzime: `CLEANUP` is where the Mender's state scripts/artifact scripts go. And it's a reasonable state for a failed download, right? Our default UM doesn't need to do any cleanup, but it's not a generic truth about all possible UMs.

danielskinstad: I don't know if I get what you're saying :sweat_smile: Are you saying it _does_ make sense to go to cleanup, but that we should not rely on the default transition since there might not be an UM, and instead e.g. hardcode transition to cleanup in this case?

vpodzime: Oh, I walked on some thin ice myself. :grin: Anyway, to clarify what I have in mind:
- an UM should have a chance to do cleanup in case of a failed download
- we need to handle missing UM (selection) carefully

So I think I actually partially agree with your change in the end. Sorry for a long way through my head. I think we need to:
- go to the cleanup state if we downloaded enough to choose the UM and esp. if we passed any data to it
- skip all states in case the download failed early before we could choose the UM
- do proper cleanup of our stuff in both cases.

Having said that, if that is way too complicating for the integration tests, we can try to assume we will handle missing UM in all cleanup/failure states gracefully. Esp. if we have tests proving that.

danielskinstad: I believe the easiest/best approach wrt the state machine tests would be for all the transitions to somehow end up in `MENDER_UPDATE_STATE_END`, and thereby know that the state machine is done with the transitions.

>  we can try to assume we will handle missing UM in all cleanup/failure states gracefully

I think this is the easiest. 
Shouldn't that be covered by what I did here, i.e. just checking if the update module is NULL when entering cleanup? Once we get the tests up and running we will hopefully prove this, as you say.

vpodzime: > Shouldn't that be covered by what I did here, i.e. just checking if the update module is NULL when entering cleanup?

Yes, in the current codebase. But I'm afraid we cannot consider the current list of states and transitions between them as set in stone. It's the future changes I'm concerned about. But, of course, if we test things properly and keep the tests up to date with future changes, there should be no problem.

lluiscampos: I agree with what I interpret as a conclusion in this thread:

That the code is good to go, the change is good, and what we have to make sure tests cover the case where there is no enough of the data downloaded so that the update module is NULL :ok_hand: 

My approval is with that in mind :point_up:

lluiscampos (approved):

PR #159, opened 2025-02-28
chore: make max state data counter configurable

danielskinstad:
I mainly want this in order to exit earlier when doing the state transition tests, but I think it makes sense to let it be configurable anyways.

### DIFF
--- a/target/zephyr/Kconfig
+++ b/target/zephyr/Kconfig
@@ -387,6 +387,14 @@ if MENDER_MCU_CLIENT
 
         menu "Error detection and recovery"
 
+           config MENDER_MAX_STATE_DATA_STORE_COUNT
+                int "Maximum number of state transitions before detecting a state loop during a deployment"
+                range 1 255
+                default 28
+                help
+                    If this limit is reached, a state loop is detected, and the deployment will be marked
+                    as a failure.
+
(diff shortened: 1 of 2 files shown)


target/zephyr/Kconfig
```diff
@@ -387,6 +387,13 @@ if MENDER_MCU_CLIENT
 
         menu "Error detection and recovery"
 
+            config MENDER_MAX_STATE_DATA_STORE_COUNT
+                int "Max state data store count"
+                range 1 255
+                default 28
+                help
+                    The maximum numbers of state transitions in a deployment.
```
vpodzime, line 395: number~s~ of...

target/zephyr/Kconfig
```diff
@@ -387,6 +387,13 @@ if MENDER_MCU_CLIENT
 
         menu "Error detection and recovery"
 
+            config MENDER_MAX_STATE_DATA_STORE_COUNT
+                int "Max state data store count"
```
vpodzime, line 391: This doesn't really reflect what the `help` says and can be quite confusing. I think keeping the name of the macro is fine, but this description needs a better wording.

vpodzime (commented):
Looks good to me otherwise.

vpodzime (approved):
Thanks for adjusting!

PR #160, opened 2025-03-04
MEN-7598: Inventory API rework

vpodzime:


vpodzime:
Tests require https://github.com/mendersoftware/mender-mcu-integration/pull/53

vpodzime:
I still need to test this manually on a board and use something like uptime for dynamic inventory.

include/mender/inventory.h
```diff
@@ -26,14 +26,31 @@ extern "C" {
...
+ */
+typedef const mender_err_t *(MenderInventoryCallback)(mender_keystore_t **inventory, uint8_t *inventory_len);
+
 #ifdef CONFIG_MENDER_CLIENT_INVENTORY
 
 /**
- * @brief Set mender inventory
- * @param inventory Mender inventory key/value pairs table, must end with a NULL/NULL element, NULL if not defined
+ * @brief Add mender inventory callback
+ * @param callback   A function to call to obtain inventory information
```
larsewi, line 42:
```suggestion
 * @param callback A function to call to obtain inventory information
```

src/core/inventory.c
```diff
@@ -181,13 +268,67 @@ mender_inventory_work_function(void) {
...
     }
 
+    /* Gather persistent inventory once */
+    if (!persistent_pushed && (NULL == persistent_inventory)) {
+        if (MENDER_OK != (ret = collect_persistent_inventory())) {
+            goto END;
+        }
+    }
+
+    /* Check there's anything to do */
```
larsewi, line 278:
```suggestion
    /* Check if there's anything to do */
```

larsewi (approved):
Very nice 🚀

src/core/api.c
```diff
@@ -667,22 +663,9 @@ mender_api_publish_inventory_data(mender_keystore_t *inventory) {
     }
     cJSON_AddStringToObject(item, "name", "device_type");
     cJSON_AddStringToObject(item, "value", api_config.device_type);
```
lluiscampos, line 665: Just a heads-up (or a double-checking),

In the next task, the idea is that `artifact_name` and `device_type` are default inventory callbacks or built-in callbacks, that they can be disabled by the user. This is mimicking the classic client functionality, were we install a set of inventory scripts by default but can be opted out by users

src/core/inventory.c
```diff
@@ -156,18 +191,70 @@ mender_inventory_exit(void) {
...
     DESTROY_AND_NULL(mender_os_mutex_delete, mender_inventory_mutex);
 
     return ret;
 }
 
 static mender_err_t
-mender_inventory_work_function(void) {
+collect_persistent_inventory(void) {
+    for (uint8_t idx = 0; idx < n_callbacks; idx++) {
+        if (callbacks[idx].persistent) {
```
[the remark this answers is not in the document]
lluiscampos: Or... did we messed up `n_persistent_callbacks` and `n_persistent_cbs` ? That might be the intention

src/core/inventory.c
```diff
@@ -181,13 +268,67 @@ mender_inventory_work_function(void) {
...
 
+    /* Gather persistent inventory once */
+    if (!persistent_pushed && (NULL == persistent_inventory)) {
+        if (MENDER_OK != (ret = collect_persistent_inventory())) {
+            goto END;
+        }
+    }
+
+    /* Check there's anything to do */
+    if ((persistent_pushed || (NULL == persistent_inventory)) && (0 == n_callbacks - n_persistent_cbs)) {
```
lluiscampos, line 279: Relates to my comment(s) above, `n_persistent_cbs` is never updated.

lluiscampos (commented):
I think there is something odd in the main logic for `mender_inventory_work_function`. Please see my comment(s) below and then I'll finish my review.

src/core/inventory.c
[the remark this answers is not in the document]
vpodzime: Good catch! You're right! I originally counted them in a look inside this function and then realized we can track it while adding them.

src/core/api.c
vpodzime: Yup, I know. And I take it this will be a persistent callback/inventory, right? Do we need to have one callback for each or can both be added by a single callback?

danielskinstad (approved):
Looks good to me. Just fyi, you don't need a `Changelog` entry for chore-commits.

vpodzime:
> Just fyi, you don't need a `Changelog` entry for chore-commits.

I know, but it's easier for me to have this in my commit message template. :)

src/core/api.c
lluiscampos: Single callback, together with the rest of the device provides.

PR #161, opened 2025-03-05
chore: log client states with names instead of numbers

danielskinstad:
The logging will go from:
`Inside work function [state: 1]`
to
`Inside work function [state: MENDER_CLIENT_STATE_OPERATIONAL]`

### DIFF
--- a/src/core/client.c
+++ b/src/core/client.c
@@ -112,6 +112,12 @@ static const char *update_state_str[N_MENDER_UPDATE_STATES + 1] = {
     "MENDER_UPDATE_STATE_FAILURE",
     "MENDER_UPDATE_STATE_END",
 };
+static const char *client_state_str[N_MENDER_CLIENT_STATES + 1] = {
+    "MENDER_CLIENT_STATE_INITIALIZATION",
+    "MENDER_CLIENT_STATE_OPERATIONAL",
+    "MENDER_CLIENT_STATE_PENDING_REBOOT",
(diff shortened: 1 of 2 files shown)


src/core/client.c
```diff
@@ -112,6 +112,12 @@ static const char *update_state_str[N_MENDER_UPDATE_STATES + 1] = {
     "MENDER_UPDATE_STATE_FAILURE",
     "MENDER_UPDATE_STATE_END",
 };
+static const char *client_state_str[N_MENDER_CLIENT_STATES + 1] = {
+    "MENDER_CLIENT_STATE_INITIALIZATION",
+    "MENDER_CLIENT_STATE_OPERATIONAL",
+    "MENDER_CLIENT_STATE_PENDING_REBOOT",
+};
```
vpodzime, line 119: Please make this debug-log-level-only.

vpodzime (approved):
Looks good to me otherwise.

src/core/client.c
danielskinstad: Is that not what it is now? `#if CONFIG_MENDER_LOG_LEVEL >= MENDER_LOG_LEVEL_DBG`? It's the same as for the `update_state_str`

vpodzime: It is, indeed! I overlooked it above. Sorry for the noise!

vpodzime (approved):

PR #162, opened 2025-03-06
chore: add server-options to Kconfig

danielskinstad:
You can now easily select which Mender server to use in the Kconfig, by choosing between:

* hosted Mender EU
* hosted MENDER US
* on-prem

Ticket: MEN-8080

### DIFF
--- a/target/zephyr/Kconfig
+++ b/target/zephyr/Kconfig
@@ -35,11 +35,39 @@ if MENDER_MCU_CLIENT
 
     menu "General configuration"
 
-        config MENDER_SERVER_HOST
-            string "Mender server host URL"
-            default "https://hosted.mender.io"
+        choice MENDER_SERVER_HOST
+            prompt "Select which Mender server to use (hosted US, hosted EU, on-premise)"
+            default MENDER_SERVER_HOST_US
             help
                 Set the Mender server host URL to be used on the device.
+
+            config MENDER_SERVER_HOST_US
+                bool "https://hosted.mender.io"
+                help
+                    hosted Mender US
+            config MENDER_SERVER_HOST_EU
+                bool "https://eu.hosted.mender.io"
+                help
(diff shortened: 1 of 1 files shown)


target/zephyr/Kconfig
```diff
@@ -35,12 +35,42 @@ if MENDER_MCU_CLIENT
...
+                bool "On-premise server"
+                help
+                    Set the Mender server host URL to be used on the device.
+            if MENDER_SERVER_HOST_ON_PREM
+                config MENDER_SERVER_HOST
+                    string "Mender server URL"
+                    default "..."
+                    help
+                        Specify a Mender server URL.
+            endif
```
vpodzime, line 62: Not sure what the difference between `depends` and `if-endif` are exactly, but I believe we use `depends` for things like this.

vpodzime (commented):
Looks good to me otherwise.

target/zephyr/Kconfig
```diff
@@ -35,12 +35,42 @@ if MENDER_MCU_CLIENT
 
     menu "General configuration"
 
-        config MENDER_SERVER_HOST
-            string "Mender server host URL"
-            default "https://hosted.mender.io"
+        choice MENDER_SERVER_HOST
+            prompt "Select which Mender server to use (EU, US, on-premise)"
```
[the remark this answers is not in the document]
danielskinstad: Good catch

target/zephyr/Kconfig
danielskinstad: I updated it with a depends, but when setting the on-prem url, the previous value is saved, and you need to erase it. Lluis` suggestion keeps the three dots, which I think is a bit nicer :+1:

vpodzime (approved):
ACK. Cool!

lluiscampos (approved):

PR #164, opened 2025-03-09
chore: always log which state we enter

danielskinstad:
I think this was suggested to me when I added the log to the download state, but nevertheless, this should ensure that we always get the `Entering state...` logs which the state-machine tests are created around


vpodzime (approved):
ACK

PR #166, opened 2025-03-10
chore: don't set default server host to `...`

danielskinstad:
I forgot to remove the `default "..."` in the previous PR, resulting in the server host being set to `...` by default


src/core/client.c
danielskinstad, line 1: My previous `fix` was always logging `Entering state...` by adding it above the switch and removing it from the failure in the macro. However, this resulted in breaking the tests, as it would log `MENDER_UPDATE_STATE_CLEANUP` twice; once in the macro, and once before entering the switch. I think this is the easiest way to do it without over-complicating things

vpodzime (approved):
ACK

PR #169, opened 2025-03-11
MEN-7687: Inventory provides

vpodzime:


src/core/inventory.c
```diff
@@ -386,4 +403,58 @@ artifact_name_device_type_cb(mender_keystore_t **inventory, uint8_t *inventory_l
...
+       need a key-value array so we need to do a transformation. */
+    if (MENDER_OK != (ret = mender_storage_get_provides(&provides))) {
+        mender_log_error("Failed to get provides from storage");
+        return ret;
+    }
+
+    n_provides = mender_utils_key_value_list_length(provides);
+    if (n_provides > UINT8_MAX) {
+        mender_log_error("Too many provides values, cannot inventory them");
+        ret = MENDER_FAIL;
```
lluiscampos, line 425: Could we truncate it instead?

(related, but I didn't check in the previous to the previous PR: if an inventory callback fails will the rest of the inventory still be submitted?)

src/include/utils.h
```diff
@@ -128,6 +128,12 @@ void mender_utils_keystore_delete(mender_keystore_t *keystore, uint8_t keystore_
  */
 mender_err_t mender_utils_key_value_list_free(mender_key_value_list_t *list);
 
+/**
+ * @brief Get linked list length
+ * @param list List to get length of
```
lluiscampos, line 133:
```suggestion
 * @param list List to get the length of
```

lluiscampos (approved):
Thank you :rocket:

src/core/inventory.c
vpodzime: > Could we truncate it instead?

Sure, I just wasn't sure about the prioritization. Clearly that `rootfs-image....` item is quite important, for example. But we can deal with that later.


> if an inventory callback fails will the rest of the inventory still be submitted?

Yes, it's a best-effort logic.

PR #170, opened 2025-03-12
fix: Cache `"unknown"` as artifact name

vpodzime:
In case the artifact name is not known (e.g. the initial state of a device until the first successful deployment), the string `"unknown"` is returned. We need to cache this value to fix two issues:

1. Make it cached and avoid trying to read it and fail again and again and
2. avoid memory leaks when the callers of the `mender_storage_get_artifact_name()` function get a `const` pointer and don't `free()` it

Ticket: MEN-8056
Changelog: none


vpodzime:
This causes a double-free.

danielskinstad (approved):
Looks good to me

PR #172, opened 2025-03-12
ci: update coverage job

danielskinstad:
Move away from the deprecated `cpp-coveralls`, and use coveralls own tool `coveralls-reporter`.

Make the report parallel so we can combine reports from mender-mcu-integration

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -8,6 +8,7 @@ stages:
   - test
   - trigger
   - publish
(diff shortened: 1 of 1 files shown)


danielskinstad:
Tested together with https://github.com/mendersoftware/mender-mcu-integration/pull/65

vpodzime (approved):
Looks good to me and seems to be working as well as coveralls can. :)

PR #175, opened 2025-03-14
chore: fail deployment on spontaneous reboot

danielskinstad:
See motivation for this change:
- https://docs.mender.io/artifact-creation/state-scripts#power-loss

A spontaneous reboot / power loss is detected when the stored state is something other than `MENDER_UPDATE_STATE_VERIFY_REBOOT` or `MENDER_UPDATE_STATE_ROLLBACK_VERIFY_REBOOT`.

Ticket: MEN-8197


src/core/client.c
```diff
@@ -1068,6 +1081,12 @@ mender_client_update_work_function(void) {
 
             case MENDER_UPDATE_STATE_INSTALL:
                 mender_log_info("Download done, installing artifact");
+                if (spontaneous_reboot) {
+                    mender_log_error("Failing deployment, spontaneous reboot detected");
+                    spontaneous_reboot = false;
+                    ret                = MENDER_FAIL;
+                    NEXT_STATE;
+                }
```
vpodzime, line 1089: What about the `REBOOT` state?

src/core/client.c
```diff
@@ -561,6 +565,15 @@ mender_client_initialization_work_function(void) {
         }
     }
 
+    /* Check for a spontaneous reboot */
+    mender_update_state_t update_state;
+    if (MENDER_OK == (ret = mender_deployment_data_get_state(mender_client_deployment_data, &update_state))) {
+        if ((MENDER_UPDATE_STATE_VERIFY_REBOOT != update_state) && (MENDER_UPDATE_STATE_ROLLBACK_VERIFY_REBOOT != update_state)) {
```
vpodzime, line 571: `DOWNLOAD` state should be ok, right?

vpodzime (commented):
Generally I don't think it's a good idea to use a negative condition in one place (not one of these two states) and then add handling of the case based on a positive condition (is one of these states). I'd prefer using the same condition in both places -- client initialization and the work function. And I'd probably even do it before the while-switch block.

src/core/client.c
danielskinstad: The docs say `Power loss is allowed to happen within any Reboot state, since it may be indistinguishable from a normal reboot`

vpodzime: Sure. But the negative check above will set `spontaneous_reboot = true` in case it boots into the `REBOOT` state and then the value is ignored. I find that confusing. That's exactly what I meant by having the negative and positive conditions for (not-exactly-)the same thing.

danielskinstad: Right, so just to get it straight, you mean that I should check that we're in DOWNLOAD or INSTALL in initialization, instead of checking that we're not in the reboot states.

vpodzime: > DOWNLOAD or INSTALL in initialization
`INSTALL`? I believe the `DOWNLOAD` state together with the `REBOOT` state**s** (incl. the `VERIFY_REBOOT_` ones) are the only ones expected. All the others mean a spontaneous reboot, right? And the same list of states should be used in both places -- when detecting a spontaneous reboot and when handling it. (potentially with a `!(...)` if that makes more sense in the code, of course)

danielskinstad: > INSTALL?

Oops, I meant `INSTALL` and `COMMIT`; I think those two are the only states that needs separate handling on a spontaneous reboot. You can double check the documentation to see if I'm missing something.

The error states: 
- `If a power loss happens inside one of the error states, that state will be repeated until it succeeds without a power interruption`

vpodzime: I see. Please make sure to add those bits from the docs as a comment somewhere in the code. Thanks for your patience!

src/core/client.c
```diff
@@ -1014,6 +1028,12 @@ mender_client_update_work_function(void) {
         continue;                                                              \
     }
 
+    if (spontaneous_reboot) {
+        mender_log_error("Failing deployment, spontaneous reboot detected");
+        spontaneous_reboot = false;
+        update_state       = update_state_transitions[update_state].failure;
+        mender_log_debug("Entering state %s", update_state_str[update_state]);
```
vpodzime, line 1035: Makes me wonder if we need `set_and_store_state()` here. I guess we don't as another reboot in the particular `failure` state would just jump here again. So there would be no difference, the `while` loop below would start with the same state in both cases.

vpodzime (approved):
Looks good to me. Thanks for reworking this!

src/core/client.c
danielskinstad: >  I guess we don't as another reboot in the particular failure state would just jump here again. So there would be no difference, the while loop below would start with the same state in both cases.

Yeah, I agree - I don't think it's necessary.

PR #179, opened 2025-03-24
Fix rollback logic

danielskinstad:


src/core/utils.c
```diff
@@ -175,6 +175,7 @@ mender_utils_vasprintf(char **result, const char *fmt, va_list ap) {
     }
     *result = mender_malloc((size_t)len + 1);
     if (NULL == *result) {
+        mender_log_error("Unable to allocate memory");
```
vpodzime, line 178: We generally don't do this in the allocation functions. And while it would have its benefits (maybe we should? :thinking: ), it's better to be consistent.

src/platform/update_modules/zephyr/image/update-module.c
```diff
@@ -298,6 +302,14 @@ mender_zephyr_image_abort_deployment(MENDER_NDEBUG_UNUSED mender_update_state_t
...
     }
     artifact_had_payload = false;
+
+    int rc;
+    if (boot_is_img_confirmed()) {
+        if ((rc = boot_erase_img_bank(MENDER_SLOT_PARTITION_1)) < 0) {
+            mender_log_error("Failed to erase second slot %d", rc);
+            return MENDER_FAIL;
+        }
+    }
```
vpodzime, line 318: This looks to me like something we should have in a `ROLLBACK` callback.

src/platform/update_modules/zephyr/image/update-module.c
```diff
@@ -298,6 +302,14 @@ mender_zephyr_image_abort_deployment(MENDER_NDEBUG_UNUSED mender_update_state_t
         return ret;
     }
     artifact_had_payload = false;
+
+    int rc;
+    if (boot_is_img_confirmed()) {
+        if ((rc = boot_erase_img_bank(MENDER_SLOT_PARTITION_1)) < 0) {
+            mender_log_error("Failed to erase second slot %d", rc);
```
vpodzime, line 315: Can we get some explanatory message for the error code? Otherwise, let's at least put it in some parentheses/brackets.

src/platform/update_modules/zephyr/image/update-module.c
```diff
@@ -326,6 +332,17 @@ mender_zephyr_image_reboot_callback(MENDER_NDEBUG_UNUSED mender_update_state_t s
     }
 }
 
+static mender_err_t
+mender_zephyr_image_rollback_callback(mender_update_state_t state, mender_update_state_data_t callback_data) {
+    assert(MENDER_UPDATE_STATE_ROLLBACK == state);
+    // Don't rollback if we're in a confirmed image - e.g. if an aborted deployment is detected in `MENDER_UPDATE_STATE_REBOOT`
+    if (mender_flash_is_image_confirmed()) {
```
vpodzime, line 339: IOW, this means there's nothing to rollback right?

src/core/utils.c
danielskinstad: We do this a lot as far as I can tell (?). I added it because I stumbled upon it while having the heap size at 8

src/platform/update_modules/zephyr/image/update-module.c
danielskinstad: Yes

src/platform/update_modules/zephyr/image/update-module.c
danielskinstad: As far as I can tell we need it here as well. If the swap type is set in INSTALL, then a reboot after a failed deployment will still attempt to boot into the image in slot 1

danielskinstad: E.g.: `INSTALL` -> `FAILURE`. If we now perform a reboot, the swap type is still set to `test`, and it will attempt to boot into the image in slot 1.

src/core/utils.c
vpodzime: Not in `mender_malloc()`, `mender_strdup()` and functions like these. We just return `NULL` from these and let the calling code handle the OOM situation.

vpodzime: I think the missing bit here is to log this message in `header_alloc_and_add()` if `mender_utils_vasprintf()` returns `<= 0`.

danielskinstad: > Not in mender_malloc(), mender_strdup() and functions like these. 

Ah, I see. I modified the `Unable to allocate memory` error then

lluiscampos (approved):

src/platform/update_modules/zephyr/image/update-module.c
vpodzime: Then, I think, we have a buggy implementation of the `INSTALL` state. Because if it actually does install, i.e. set the swap type to `test`, it needs to make sure that **before** that the state is updated to `REBOOT` so that the rollback path is followed.

danielskinstad:
This doesn't work properly yet, amongst other the current implementation increases the state counter twice

src/core/client.c
```diff
@@ -1110,14 +1110,27 @@ mender_client_update_work_function(void) {
                 /* Check ret to see if the deployment is aborted */
                 ret = mender_client_publish_deployment_status(deployment_id, MENDER_DEPLOYMENT_STATUS_INSTALLING);
                 if ((MENDER_ABORTED != ret) && (NULL != mender_update_module->callbacks[update_state])) {
+                    if (mender_update_module->requires_reboot) {
+                        /* Save the next state before running the install callback.
+                         * If there is a spontaneous reboot we assure that we don't
```
vpodzime, line 1115: *...a spontaneous reboot right after the callback we assure...*

src/core/client.c
```diff
@@ -1110,14 +1110,27 @@ mender_client_update_work_function(void) {
                 /* Check ret to see if the deployment is aborted */
                 ret = mender_client_publish_deployment_status(deployment_id, MENDER_DEPLOYMENT_STATUS_INSTALLING);
                 if ((MENDER_ABORTED != ret) && (NULL != mender_update_module->callbacks[update_state])) {
+                    if (mender_update_module->requires_reboot) {
+                        /* Save the next state before running the install callback.
```
vpodzime, line 1114: The commit message should also mention it is the INSTALL callback in the title.

vpodzime (approved):
Just nitpicks. Thanks!

src/platform/update_modules/zephyr/image/update-module.c
```diff
@@ -298,6 +303,14 @@ mender_zephyr_image_abort_deployment(MENDER_NDEBUG_UNUSED mender_update_state_t
         return ret;
     }
     artifact_had_payload = false;
+
+    int rc;
+    if (boot_is_img_confirmed()) {
+        if ((rc = boot_erase_img_bank(MENDER_SLOT_PARTITION_1)) < 0) {
+            mender_log_error("Failed to erase second slot (result = %d, errno = %s)", rc, strerror(errno));
```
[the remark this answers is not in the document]
danielskinstad: Same here; it doesn't seem to work, it prints `%m`

PR #180, opened 2025-03-24
fix: return `MENDER_FAIL` if authentication fails

danielskinstad:
Check the return value from `ensure_authenticated_and_locked()` and return `MENDER_FAIL` on failures to authenticate.

Ticket: MEN-8219
Changelog: Title

### DIFF
--- a/src/core/api.c
+++ b/src/core/api.c
@@ -127,12 +127,13 @@ mender_api_ensure_authenticated(void) {
         /* Error already logged. */
         return MENDER_FAIL;
     }
+    bool authenticated = ((MENDER_OK == ret) || (MENDER_DONE == ret));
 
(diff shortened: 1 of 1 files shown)


src/core/api.c
```diff
@@ -127,12 +127,13 @@ mender_api_ensure_authenticated(void) {
         /* Error already logged. */
         return MENDER_FAIL;
     }
+    bool authenticated = (MENDER_OK == ret);
```
vpodzime, line 130: `MENDER_DONE` also means we are authenticated, right?

danielskinstad: Right, nice catch

lluiscampos (approved):
Good catch

vpodzime (approved):
ACK. Thanks!

PR #183, opened 2025-03-26
[MEN-7626] Feat: Deployment logs

vpodzime:


src/platform/storage/zephyr/nvs/storage.c
```diff
@@ -159,6 +159,9 @@ mender_storage_init(void) {
         mender_log_error("Unable to mount NVS storage, result = %d", result);
         return MENDER_FAIL;
     }
+    mender_log_debug("Initialized Mender NVS with %u sectors (%zd bytes available)",
```
lluiscampos, line 196: `sector.size` is `uint16_t`, not `size_t`, so this should be `%u bytes` (?)

src/include/storage.h
```diff
@@ -82,6 +82,38 @@ mender_err_t mender_storage_get_deployment_data(char **deployment_data);
...
+ * @param msg_size Size of the message (length + 1)
+ * @return MENDER_OK if the function succeeds, error code otherwise
+ */
+mender_err_t mender_storage_deployment_log_append(const char *msg, size_t msg_size);
+
+/**
+ * @brief Visitor function for deployment log messages
+ * @param msg Deployment log message
+ * @param ctx Arbitrary context data
+ * @return MENDER_OK if the function succeeds, error code otherwise
```
lluiscampos, line 98: Returns void

src/platform/storage/zephyr/nvs/storage.c
```diff
@@ -52,6 +57,24 @@ static char *cached_artifact_name = NULL;
  */
 static struct nvs_fs mender_storage_nvs_handle;
 
+#ifdef CONFIG_MENDER_DEPLOYMENT_LOGS
+/**
+ * @brief A Flash Circular Buffer for storing deployment logs
+ */
+static struct fcb depl_logs_buffer;
+
+#define DEPL_LOGS_MAX_MSG_LEN 255
```
lluiscampos, line 66: Could we have this parameter configured by the user (Kconfig)? Just asking, good to know for future extensions.

src/platform/storage/zephyr/nvs/storage.c
```diff
@@ -136,22 +159,33 @@ crc_check(const unsigned char *data, const size_t data_len) {
...
+    n_sectors = sizeof(flash_sectors) / sizeof(flash_sectors[0]);
+    result    = flash_area_get_sectors(MENDER_STORAGE_FLASH_AREA, &n_sectors, flash_sectors);
+    if ((0 != result) && (-ENOMEM != result)) {
+        /* -ENOMEM means there were more sectors than the supplied size of the
+            sector array (flash_sectors), but we don't worry about that, we just
+            need the info about the first N sectors we want to use. */
+        mender_log_error("Failed to get info about flash sectors in the Mender flash area [%d]", -result);
+        return MENDER_FAIL;
+    }
+    if (n_sectors != (sizeof(flash_sectors) / sizeof(flash_sectors[0]))) {
```
lluiscampos, line 181: You know much better this stuff, but this comparison makes me wary: can we compare unsigned int to `size_t/size_t`?

Edit: I asked ChatGPT and it said it is safe :robot:

src/platform/storage/zephyr/nvs/storage.c
```diff
@@ -163,9 +197,149 @@ mender_storage_init(void) {
                      mender_storage_nvs_handle.sector_count,
                      (mender_storage_nvs_handle.sector_count - 1) * mender_storage_nvs_handle.sector_size);
 
+#ifdef CONFIG_MENDER_DEPLOYMENT_LOGS
+    /* Initialize the Flash Circular Buffer (FCB) for deployment logs. */
+
+    /* just some fixed magic value the FCB implementation uses to mark
+       in-use/erased/... sectors */
+    depl_logs_buffer.f_magic       = 0x2AABEE35;
```
lluiscampos, line 205: The magic seems to be completely user defined. Consider using a Kconfig setting for it

src/platform/storage/zephyr/nvs/storage.c
```diff
@@ -163,9 +197,149 @@ mender_storage_init(void) {
...
+        return MENDER_FAIL;
+    }
+
+    struct fcb_entry entry = { 0 };
+    result                 = fcb_append(&depl_logs_buffer, msg_size, &entry);
+    if (-ENOSPC == result) {
+        if (0 != (result = fcb_rotate(&depl_logs_buffer))) {
+            mender_log_error("Failed to rotate the deployment logs FCB");
+            /* if (0 != (result = fcb_append_to_scratch(&depl_logs_buffer))) { */
+            /*     mender_log_error("Failed to start using the scratch sector for the deployment logs FCB"); */
```
lluiscampos, line 261: Clean-up :broom:

src/platform/storage/zephyr/nvs/storage.c
```diff
@@ -163,9 +197,149 @@ mender_storage_init(void) {
...
+
+    const struct flash_area *fap;
+    if (0 != (result = flash_area_open(FIXED_PARTITION_ID(MENDER_STORAGE_LABEL), &fap))) {
+        mender_log_error("Unable to open the Mender storage flash area");
+        return MENDER_FAIL;
+    }
+
+    struct fcb_entry entry = { 0 };
+    result                 = fcb_append(&depl_logs_buffer, msg_size, &entry);
+    if (-ENOSPC == result) {
```
lluiscampos, line 257: All this handling could be done on a separate static function. It would just read better :eyes:

src/platform/storage/zephyr/nvs/storage.c
```diff
@@ -163,9 +197,149 @@ mender_storage_init(void) {
...
+    }
+
+END:
+    flash_area_close(fap);
+
+    return ret;
+}
+
+mender_err_t
+mender_storage_deployment_log_walk(MenderDeploymentLogVisitor visitor_fn, void *ctx) {
```
lluiscampos, line 306: It is the first time I see the visitor pattern in C :heart_eyes:

src/platform/storage/zephyr/nvs/storage.c
```diff
@@ -163,9 +197,149 @@ mender_storage_init(void) {
...
+        return MENDER_FAIL;
+    }
+
+    struct fcb_entry entry = { 0 };
+    result                 = fcb_getnext(&depl_logs_buffer, &entry);
+    while (0 == result) {
+        if (0 != (result = flash_area_read(fap, entry.fe_sector->fs_off + entry.fe_data_off, (void *)msg, MIN(DEPL_LOGS_MAX_MSG_LEN, entry.fe_data_len)))) {
+            mender_log_error("Failed to read FCB entry from flash [%d]", -result);
+            ret = MENDER_FAIL;
+            goto END;
```
lluiscampos, line 323: In this case, why not `break`? I guess we are just trying to be future proof?

src/platform/storage/zephyr/nvs/storage.c
```diff
@@ -163,9 +197,149 @@ mender_storage_init(void) {
...
+        }
+        msg[MIN(DEPL_LOGS_MAX_MSG_LEN, entry.fe_data_len)] = '\0';
+        visitor_fn(msg, ctx);
+        result = fcb_getnext(&depl_logs_buffer, &entry);
```
lluiscampos, line 327: If `fcb_getnext` gives any error it will break the loop. The intention is to break on "no more entries" error, but we would be masking any other error too.

We should check outside of the loop that result is "no more entries" error.

target/zephyr/Kconfig
```diff
@@ -86,6 +86,13 @@ if MENDER_MCU_CLIENT
             help
                 Interval used to periodically check for new deployments on the Mender server.
 
+        config MENDER_DEPLOYMENT_LOGS
+            bool "Deployment logs"
+            default y
+            select FCB
+            help
+                Enable the deployment logs feature.
```
lluiscampos, line 94: I would extend here a bit more the help text. For example:

This is a feature that saves in non-volatile storage the Mender MCU warning and error logs during a deployment so that they can be send to the Mender Server in case on a failed deployment for further diagnostics. It incurs flash wear.

src/include/log.h
```diff
@@ -38,6 +38,20 @@ mender_err_t mender_log_init(void);
  */
 mender_err_t mender_log_exit(void);
 
+#ifdef CONFIG_MENDER_DEPLOYMENT_LOGS
+/**
+ * @brief Activate deployment logs gathering
```
lluiscampos, line 43: General comment,

I think "store" or "save" is better than "gather". (If you decide to change, do so both in the docs and in the log messages)

src/platform/log/zephyr/log.c
```diff
@@ -40,3 +82,76 @@ mender_log_exit(void) {
...
+    if (msg->log.hdr.desc.level > LOG_LEVEL_WRN) {
+        return;
+    }
+
+    uint32_t          flags           = log_backend_std_get_flags() & ~LOG_OUTPUT_FLAG_COLORS;
+    log_format_func_t log_format_func = log_format_func_t_get(LOG_OUTPUT_TEXT);
+
+    log_format_func(&depl_log_output, &msg->log, flags);
+}
+#endif /* CONFIG_MENDER_DEPLOYMENT_LOGS */
```
lluiscampos, line 157: All in all the deployment logs code is around 75% of the file now, and there is nothing shared with the rest of the "normal" logging. Consider moving it to `log_deployment.c`.

src/platform/log/zephyr/log.c
```diff
@@ -40,3 +82,76 @@ mender_log_exit(void) {
...
+}
+
+static void
+process(MENDER_ARG_UNUSED const struct log_backend *const backend, union log_msg_generic *msg) {
+    /* We only care about warning and error log messages */
+    if (msg->log.hdr.desc.level > LOG_LEVEL_WRN) {
+        return;
+    }
+
+    uint32_t          flags           = log_backend_std_get_flags() & ~LOG_OUTPUT_FLAG_COLORS;
```
lluiscampos, line 152: Nice find :mag_right:

src/core/api.c
```diff
@@ -616,6 +621,204 @@ mender_api_publish_deployment_status(const char *id, mender_deployment_status_t
...
+        }
+    }
+#endif /* CONFIG_MENDER_DEPLOYMENT_LOGS */
+
+    return ret;
+}
+
+#ifdef CONFIG_MENDER_DEPLOYMENT_LOGS
+static void
+append_depl_log_msg(char *msg, void *ctx) {
```
lluiscampos, line 649: This function we should unit test

lluiscampos, line 649: Why `void *`? There is only one caller and no magic needed I think.

src/core/api.c
```diff
@@ -616,6 +621,204 @@ mender_api_publish_deployment_status(const char *id, mender_deployment_status_t
...
+        /* skip the space */
+        c++;
+    }
+
+    if ('\0' != *c) {
+        log_msg = c;
+        if (mender_utils_strbeginswith(log_msg, "mender: ")) {
+            log_msg += strlen("mender: ");
+        }
+    }
```
lluiscampos, line 713: I haven't reviewed the parsing, sorry. A unit tests will build up my confidence :)

lluiscampos (commented):
It looks very good :100:

src/platform/log/zephyr/log.c
```diff
@@ -40,3 +82,76 @@ mender_log_exit(void) {
...
+    return MENDER_OK;
+}
+
+mender_err_t
+mender_deployment_logs_deactivate(void) {
+    MENDER_NDEBUG_UNUSED mender_err_t ret;
+
+    /* In case there is an incomplete log message, save it. */
+    if (0 != msg_tmp_idx) {
+        msg_tmp[msg_tmp_idx++] = '\0';
```
larsewi, line 99: Should we append `...` or something similar to signal that it is incomplete? Maybe not worth the extra code 🤔

larsewi (commented):
Very nice 🍻🚀 

`s/DOWNLOAd/DOWNLOAD/` in 2405f5ecc50d0b043df95cc2fd0eaf847c35b1df

src/platform/storage/zephyr/nvs/storage.c
vpodzime: Do you think people would like to set it themselves? Why?

src/platform/storage/zephyr/nvs/storage.c
vpodzime: Then you should see "my" `JsonWalk()`: https://github.com/NorthernTechHQ/libntech/blob/master/libutils/json.h#L543-L568 :sweat_smile:

src/core/api.c
vpodzime: Because that's the type of the visitor function.

vpodzime:
Thanks for the reviews, guys! Good findings! I'll address some and leave some others to follow-up PRs (esp. unit tests :grin:  ).

src/platform/storage/zephyr/nvs/storage.c
lluiscampos: I don't have a reason. Is just that it looks very esoteric to have here a hard-coded value. What if we have it at the beginning of the file with a define at least?

src/platform/storage/zephyr/nvs/storage.c
```diff
@@ -163,9 +197,149 @@ mender_storage_init(void) {
...
+#ifdef CONFIG_MENDER_DEPLOYMENT_LOGS
+mender_err_t
+mender_storage_deployment_log_append(const char *msg, size_t msg_size) {
+    int          result;
+    mender_err_t ret = MENDER_OK;
+
+    msg_size = MIN(DEPL_LOGS_MAX_MSG_LEN + 1, msg_size);
+
+    const struct flash_area *fap;
+    if (0 != (result = flash_area_open(FIXED_PARTITION_ID(MENDER_STORAGE_LABEL), &fap))) {
```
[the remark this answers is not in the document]
vpodzime: Might as well log it :)

src/platform/log/zephyr/log.c
vpodzime: Yeah, I think it will be pretty obvious. We can revisit this if there's any need.

src/platform/storage/zephyr/nvs/storage.c
vpodzime: I think it's better to add `(size_t)` in front of the multiplication.

src/platform/storage/zephyr/nvs/storage.c
vpodzime: Yes, we could. My only concern is that then it could be different between different versions/images and that could cause a mismatch/misalignment. I guess we would then need to use the `version` parameter of the FCB a ensure there's some backwards compatibility. For now, I'd keep it fixed. That's also what I have seen in other logging backends I checked.

src/platform/storage/zephyr/nvs/storage.c
vpodzime: Yes, exactly. And to follow the common pattern in the codebase.

src/core/api.c
vpodzime: I agree. But have we agreed on how we'll unit test `static` functions?

lluiscampos: We don't have tests for static functions directly (not yet, at least). But since we have now private and public header files, that is a easy way to expose the bits we want to test.

src/platform/log/zephyr/log.c
vpodzime: I did consider that. But since there was (and would be again) basically nothing in `log.c`, I don't think we need that in a separate file. Perhaps in the future there's more to add to this file and the split will make more sense. But I'm leaving it as is for now.

src/core/api.c
vpodzime: > that is a easy way to expose the bits we want to test

It's not, unfortunately. It's not only about headers, you'd still have to drop `static` from the function declaration/definition and it would get into the linker namespace, potentially causing conflicts and unexpected behavior.

larsewi (approved):

src/core/api.c
lluiscampos: I meant both: removing the static and moving it into the private header (with a prefix). The point is that it is still out of the public header file so it should not be on the way for users or other compile units.

lluiscampos (approved):

src/core/api.c
vpodzime: Yeah, but then we need to prefix the name with `mender_`. And I generally don't like the idea of making functions non-static just because we want to unit-test them.

PR #186, opened 2025-03-28
MEN-7548: Make inventory "feature" hard to disable

lluiscampos:
While we want to keep the code around for potentially minimalistic builds of the Client, we at Mender team consider it a core feature that should not be opted out.

The middle point is to mark the disabling of the inventory as an EXPERIMENTAL feature (we won't be testing it, after all) but keep the code around for hackers or advance users to play with.


lluiscampos:
Goes with:
* https://github.com/mendersoftware/mender-mcu-integration/pull/73

larsewi (approved):

src/platform/inventory/zephyr/inventory.c
```diff
@@ -25,6 +25,8 @@
 #include "log.h"
 #include "utils.h"
 
+#ifndef CONFIG_MENDER_CLIENT_INVENTORY_DISABLE
```
vpodzime, line 28: This is not necessary as the whole file is omitted from the build in such a case, AFAIK. Should we check other files that might be completely depending on a config option and similar checks there?

target/zephyr/Kconfig
```diff
@@ -95,36 +95,33 @@ if MENDER_MCU_CLIENT
...
+        config MENDER_CLIENT_INVENTORY_NETWORK_INFO
+            bool "Network info built-in inventory"
+            default y
+            help
+                Enable built-in network info inventory
 
-        endif
+        config MENDER_CLIENT_INVENTORY_DISABLE
+            bool "Disable Mender client Inventory (EXPERIMENTAL)"
+            default n
```
vpodzime, line 120: Shouldn't all the above have `depend on MENDER_CLIENT_INVENTORY_DISABLE=n`?

src/platform/inventory/zephyr/inventory.c
lluiscampos: You are right, but it was there already for `src/inventory/inventory.c` and I wanted to keep it consistent. Do you prefer that I remove the other one instead?

I decided this way as it is harmless and if any, more defensive than the opposite (to remove it from `src/inventory/inventory.c`)

target/zephyr/Kconfig
lluiscampos: Technically yes, but it is harmless to let the user define them even when disabling inventory, and this being an experimental feature I didn't want to put it on top (which would make more sense) but then seems like we encourage it more.

If you insist I'll modify it.

src/platform/inventory/zephyr/inventory.c
vpodzime: We can leave the change to drop these in favor of skipping whole files to some cleanup PR. Not a biggie and as you say, it was already there. :+1:

target/zephyr/Kconfig
vpodzime: It can still be the last item and the ones above will disappear once the user selects it. They will see how much they are losing! Even more pressure! :wink:

danielskinstad (approved):

target/zephyr/Kconfig
lluiscampos: Modified

PR #187, opened 2025-04-01
Update work queue variables

danielskinstad:


vpodzime (commented):
Actually, from my side, renaming `CONFIG_MENDER_SCHEDULER_SEPARATE_WORK_QUEUE` to `CONFIG_MENDER_OS_SEPARATE_WORK_QUEUE` was a bug in my changes. I think we should keep `SCHEDULER` in the scheduler-related items.

Looks good to me otherwise.

lluiscampos (approved):

PR #188, opened 2025-04-02
fix: reschedule work items with backoff on network errors

danielskinstad:
If `MENDER_NET_ERROR` is detected in a work function it will be retried with a separate backoff interval which doubles on each try. It will continue until it reaches the max backoff interval and keep retrying using the max backoff interval.

Added configurable options for backoff:
* MENDER_NETWORK_ERROR_BACKOFF
* MENDER_NETWORK_ERROR_MAX_BACKOFF

Ticket: MEN-8222
Changelog: Title


vpodzime (commented):
I'll go through this in detail after lunch. But first two observations:
- I think `uint32_t` is unnecessarily big for the retry interval(s), it doesn't make sense to have them longer than `UINT16_MAX`, right?
- We really need a retry of the HTTP requests themselves, as the ticket says esp. for the deployment status posting (and thus checking for aborted deployments).

I can take care of the latter as part of MEN-8223, though.

src/core/api.c
```diff
@@ -256,7 +256,7 @@ perform_authentication(void) {
         mender_api_print_response_error(response, status);
         /* Maybe the identity is wrong? Let's make sure we get fresh data for the next attempt. */
         FREE_AND_NULL(identity_info);
-        ret = MENDER_FAIL;
+        ret = MENDER_NET_ERROR;
```
vpodzime, line 259: This is the case were we get a response from the server, that's not a network error. I guess re-authentication attempts should happen sooner than after an update check interval. What about changing `NET_ERROR` to `RETRY_ERROR` and use it for all cases that should retry? I'm not sure if we should retry on all bad statuses, though.

src/core/api.c
```diff
@@ -546,7 +546,7 @@ mender_api_check_for_deployment(mender_api_deployment_data_t *deployment) {
         ret = MENDER_NOT_FOUND;
     } else {
         mender_api_print_response_error(response, status);
-        ret = MENDER_FAIL;
+        ret = MENDER_NET_ERROR;
```
vpodzime, line 549: Same here and below -- if we get a response and have a status, it's not a network error.

vpodzime (commented):
Looks good to me otherwise (together with the things pointed out above earlier today).

src/core/api.c
danielskinstad: > What about changing NET_ERROR to RETRY_ERROR

I think I like this. We had this issue the c++ client; there we added a backoff on _all failed attempts to communicate with the server_. I tried to do something like that.

src/core/api.c
lluiscampos: (maybe this was discussed outside?)

Even when getting an status, we will retry with the retry interval. This is specially important for 401 Unauthorized, as you don't want to wait the full poll interval after you sent your first auth request. Other status like rate limiting or server errors should also follow this logic. This was a painful lesson learned from Mender Client 3 to Mender Client4/5, and there we ended up by on all error responses.

vpodzime:
> Does this include the retry of the https requests, or should I do that in a follow-up PR

Discussed at today's client team meeting. I'll take care of the "burst retries" in my PR and create a follow-up ticket for the longer/state-machine-based retry.

include/mender/client.h
```diff
@@ -40,7 +40,9 @@ typedef struct {
 #ifndef CONFIG_MENDER_CLIENT_INVENTORY_DISABLE
     uint32_t inventory_update_interval; /**< Inventory update interval, default is compile-time defined */
 #endif                                  /* CONFIG_MENDER_CLIENT_INVENTORY_DISABLE */
-    bool recommissioning;               /**< Used to force creation of new authentication keys */
+    uint16_t backoff_interval;          /**< Backoff interval for retries on retry errors, default is compile-time defined */
```
vpodzime, line 43: ~...default is...~ *...0 means use...*

src/platform/os/zephyr/os.c
```diff
@@ -95,8 +97,12 @@ mender_os_scheduler_work_create(mender_os_scheduler_work_params_t *work_params,
     }
 
     /* Copy work parameters */
-    work_context->params.function = work_params->function;
-    work_context->params.period   = work_params->period;
+    work_context->params.function             = work_params->function;
+    work_context->params.period               = work_params->period;
+    work_context->params.backoff.max_interval = work_params->backoff.max_interval;
+    /* Store the backoff interval so we can reset it */
+    work_context->params.backoff.interval = (backoff_interval = work_params->backoff.interval);
```
vpodzime, line 104: Please do this in two steps.

src/platform/os/zephyr/os.c
```diff
@@ -222,21 +228,34 @@ mender_os_scheduler_work_handler(struct k_work *work_item) {
...
     }
     if (MENDER_OK != ret) {
-        mender_log_error("Work %s failed", work->params.name);
+        if (MENDER_RETRY_ERROR == ret) {
+            mender_log_debug("Retry error detected, retrying with backoff");
+            period                        = work->params.backoff.interval;
+            uint16_t next                 = work->params.backoff.interval * 2;
+            work->params.backoff.interval = (next >= work->params.backoff.max_interval) ? work->params.backoff.max_interval : next;
+        }
+        mender_log_error("Work %s failed, retrying in %u seconds", work->params.name, period);
```
vpodzime, line 248: `%"PRIu32" seconds` should be used to print an `uint32_t` value.

target/zephyr/Kconfig
```diff
@@ -586,6 +586,20 @@ if MENDER_MCU_CLIENT
                 help
                     The number of iterations to wait for a reboot.
 
+            config MENDER_RETRY_ERROR_BACKOFF
+                int "Backoff interval on retry errors (seconds)"
+                default 60
```
vpodzime, line 591: I'd add a range, for two reasons:
- it cannot be a negative number and
- it doesn't make sense to set very high values here and below.

vpodzime (commented):
Looks good to me otherwise. I'll tentatively rebase #192 on top of this.

target/zephyr/Kconfig
danielskinstad: Right, I set it at 1 hour max for the starting interval and 6 hours max for the max interval (completely arbitrary). Should we set the range from 0 to the largest integer uint16_t can hold or is this fine

src/platform/net/zephyr/http.c
```diff
@@ -231,7 +231,7 @@ mender_http_perform(char                *jwt,
 
     mender_free(request.recv_buf);
 
-    return ret;
+    return (MENDER_FAIL != ret) ? ret : MENDER_RETRY_ERROR;
```
vpodzime, line 234: If this part fails:
```

    /* Perform HTTP request */
    if ((http_req_ret = http_client_req(sock, &request, MENDER_HTTP_REQUEST_TIMEOUT, (void *)&request_context)) < 0) {
        mender_log_error("HTTP request failed: %s", strerror(-http_req_ret));
        goto END;
    }
```
`ret` will be `MENDER_FAIL` because the above call doesn't modify it in any way. And that's exactly where a network error would surface/happen, I believe.

vpodzime: Bad reading of the code, sorry for the noise.

target/zephyr/Kconfig
vpodzime: I believe in real numbers! I mean, the real real ones, not the non-integer ones :grimacing:

vpodzime (approved):
Looks good to me now. Thanks!

PR #189, opened 2025-04-02
MEN-8243: Support for storage partition selection

vpodzime:


vpodzime:
I tried to make the code as readable as possible, but it's quite tricky with the complexity that is required now. :thinking:

vpodzime:
> [ci/gitlab](https://gitlab.com/Northern.tech/Mender/mender-mcu/-/pipelines/1748941444) — Pipeline failed on Gitlab

Only an unrelated failure due to a bug in the pipeline which tries to upload coverage data even when no integration tests are being run (because this change requires https://github.com/mendersoftware/mender-mcu-integration/pull/75 and thus a custom run of the pipeline with an unprotected mender-mcu-integration branch).

src/platform/storage/zephyr/nvs/storage.c
```diff
@@ -163,41 +211,85 @@ crc_check(const unsigned char *data, const size_t data_len) {
...
+        return MENDER_FAIL;
+    }
+#else
+    /* With custom storage, we need to call mender_get_storage_spec() to get the
+       sector offset and only then we'll know how many offsets we need to work
+       with. */
+    if (MENDER_OK != mender_get_storage_spec(&(mender_storage_nvs_handle.flash_device), &part_id, &sector_offset)) {
+        mender_log_error("Failed to get custom storage spec");
+        return MENDER_FAIL;
+    }
```
lluiscampos, line 241: Except for the `assert`, this block is the same as above. Consider moving it out of the `#ifndef` / `#else` and adjust a bit the `n_sectors = ...` line

src/platform/storage/zephyr/nvs/storage.c
```diff
@@ -163,41 +211,85 @@ crc_check(const unsigned char *data, const size_t data_len) {
...
+    n_sectors = sector_offset + CONFIG_MENDER_STORAGE_NVS_SECTOR_COUNT + CONFIG_MENDER_STORAGE_DEPLOYMENT_LOGS_SECTORS;
+    if (NULL == (flash_sectors = mender_calloc(sizeof(struct flash_sector), n_sectors))) {
+        mender_log_error("Cannot allocate memory");
+        return MENDER_FAIL;
+    }
+#endif /* CONFIG_MENDER_STORAGE_PARTITION_CUSTOM */
 
     /* Get flash info */
```
lluiscampos, line 249: Remove?

src/platform/storage/zephyr/nvs/storage.c
```diff
@@ -163,41 +211,85 @@ crc_check(const unsigned char *data, const size_t data_len) {
...
             sector array (flash_sectors), but we don't worry about that, we just
             need the info about the first N sectors we want to use. */
         mender_log_error("Failed to get info about flash sectors in the Mender flash area [%d]", -result);
         return MENDER_FAIL;
     }
-    if (n_sectors != (sizeof(flash_sectors) / sizeof(flash_sectors[0]))) {
-        mender_log_error(
-            "Not enough sectors on the flash for Mender (required: %" PRIu32 "d, available: %zu)", sizeof(flash_sectors) / sizeof(flash_sectors[0]), n_sectors);
+    if (n_sectors != n_sectors_required) {
+        mender_log_error("Not enough sectors on the flash for Mender (required: %" PRIu32 "d, available: %" PRIu32 "d)", n_sectors_required, n_sectors);
```
lluiscampos, line 277: Pre-existing, but the trailing `d`s here are wrong (right?). The `PRIu32` is the correct and complete format qualifier. I guess it is just printing a literal `d` at the end.

src/platform/storage/zephyr/nvs/storage.c
```diff
@@ -163,41 +211,85 @@ crc_check(const unsigned char *data, const size_t data_len) {
...
+    mender_storage_nvs_handle.sector_size  = (uint16_t)flash_sectors[sector_offset].fs_size;
     mender_storage_nvs_handle.sector_count = CONFIG_MENDER_STORAGE_NVS_SECTOR_COUNT;
 
     /* Mount NVS */
     if (0 != (result = nvs_mount(&mender_storage_nvs_handle))) {
         mender_log_error("Unable to mount NVS storage, result = %d", result);
         return MENDER_FAIL;
     }
-    mender_log_debug("Initialized Mender NVS with %u sectors (%zu bytes available)",
+    mender_log_debug("Initialized Mender NVS at 0x%jx with %u sectors (%zu bytes available)",
```
lluiscampos, line 291: Have you verified that the length modifier `j` is available?

target/zephyr/Kconfig
```diff
@@ -472,6 +472,34 @@ if MENDER_MCU_CLIENT
 
         menu "Storage options"
 
+            choice MENDER_STORAGE_PARTITION
+                prompt "Storage partition to use for Mender non-volatile data"
+                default MENDER_STORAGE_PARTITION_MENDER_PARTITION
+                help
+                    "Which storage partition (defined in the device tree) Mender should use for storing non-volatile data"
```
lluiscampos, line 479: At least in Zephyr documentation, they use the capitalization `Devicetree`

lluiscampos (commented):
It looks very good :100: There wasn't a lot that I could comment on :grin:

lluiscampos (approved):

src/platform/storage/zephyr/nvs/storage.c
```diff
@@ -163,41 +211,85 @@ crc_check(const unsigned char *data, const size_t data_len) {
...
+    n_sectors = sector_offset + CONFIG_MENDER_STORAGE_NVS_SECTOR_COUNT + CONFIG_MENDER_STORAGE_DEPLOYMENT_LOGS_SECTORS;
+    if (NULL == (flash_sectors = mender_calloc(sizeof(struct flash_sector), n_sectors))) {
+        mender_log_error("Cannot allocate memory");
+        return MENDER_FAIL;
+    }
+#endif /* CONFIG_MENDER_STORAGE_PARTITION_CUSTOM */
 
     /* Get flash info */
-    mender_storage_nvs_handle.flash_device = MENDER_STORAGE_DEVICE;
+    /** First check if the device is ready. */
```
larsewi, line 250: Why slash double star? Does it mean something special? I know we use it for doc strings.

larsewi (approved):
ACK

danielskinstad (approved):

src/platform/storage/zephyr/nvs/storage.c
vpodzime: Why? It's the beginning of the code that gets flash info.

src/platform/storage/zephyr/nvs/storage.c
vpodzime: Right! The `intN_t` man page says these are *length modifiers* which, to me, says they should be followed by a *conversion specifier* as the `printf(3)` man page says, but that's not the case. Good catch! And a reminder I should test this case manually.

src/platform/storage/zephyr/nvs/storage.c
vpodzime: Yes. And I think it's a pretty standard one, part of the `printf(3)` man page.

src/platform/storage/zephyr/nvs/storage.c
vpodzime: (update: works fine with the `d`s dropped )

src/platform/storage/zephyr/nvs/storage.c
larsewi: cool :smile:

PR #190, opened 2025-04-03
feat: added dormant certificate

danielskinstad:
Suggestion for how to add the dormant/backup certificate. I didn't think it made sense to add it to `mender-mcu-integration`. Either way, we'll create a follow-up ticket for actually testing the X2 certificate, as it turned out to be non-trivial to create a server with that specific root ca

### DIFF
--- a/cmake/CMake_posix_defaults.txt
+++ b/cmake/CMake_posix_defaults.txt
@@ -21,6 +21,7 @@ set(CONFIG_MENDER_PLATFORM_SCHEDULER_TYPE "posix" CACHE STRING "Scheduler type")
 set(CONFIG_MENDER_PLATFORM_STORAGE_TYPE "posix" CACHE STRING "Storage type")
 set(CONFIG_MENDER_PLATFORM_TLS_TYPE "generic/mbedtls" CACHE STRING "TLS type")
 set(CONFIG_MENDER_PLATFORM_SHA_TYPE "generic/mbedtls" CACHE STRING "SHA type")
+set(CONFIG_MENDER_PLATFORM_CERT_TYPE "posix" CACHE STRING "Certificate type")
 
 include(${CMAKE_CURRENT_LIST_DIR}/CMake_defaults.txt)
 
--- a/src/core/client.c
+++ b/src/core/client.c
@@ -32,6 +32,10 @@
 #include "deployment-data.h"
 #include "error-counters.h"
 
+#ifdef CONFIG_MENDER_SERVER_DORMANT_CERTIFICATE
+#include "certs.h"
+#endif /* CONFIG_MENDER_SERVER_DORMANT_CERTIFICATE */
+
 #ifndef CONFIG_MENDER_CLIENT_INVENTORY_DISABLE
 #include "inventory.h"
 #endif /* CONFIG_MENDER_CLIENT_INVENTORY_DISABLE */
@@ -352,6 +356,14 @@ mender_client_init(mender_client_config_t *config, mender_client_callbacks_t *ca
         mender_log_error("Unable to initialize API");
         goto END;
     }
+
+#ifdef CONFIG_MENDER_SERVER_DORMANT_CERTIFICATE
+    if (MENDER_OK != (ret = mender_add_dormant_cert())) {
+        /* error logged in function */
+        goto END;
+    }
+    mender_log_debug("Added dormant certificate");
+#endif
 
 #ifndef CONFIG_MENDER_CLIENT_INVENTORY_DISABLE
     if (MENDER_OK
--- a/src/include/certs.h
+++ b/src/include/certs.h
@@ -1,0 +1,39 @@
+/**
+ * @file      certs.h
+ * @brief     Mender MCU Certificate (private API)
+ *
+ * Copyright Northern.tech AS
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef __MENDER_CERTS_PRIV_H__
+#define __MENDER_CERTS_PRIV_H__
+
+#ifdef __cplusplus
+extern "C" {
+#endif /* __cplusplus */
+
+#include "utils.h"
+
+/**
+ * @brief Add dormant certificate for potential disaster recovery scenarios
+ * @return MENDER_OK if the function succeeds, error code otherwise
+ */
+mender_err_t mender_add_dormant_cert(void);
+
+#ifdef __cplusplus
+}
+#endif /* __cplusplus */
+
+#endif /* __MENDER_CERTS_PRIV_H__ */
--- a/src/platform/certs/zephyr/certs.c
+++ b/src/platform/certs/zephyr/certs.c
@@ -1,0 +1,42 @@
+/**
+ * @file      certs.c
+ * @brief     Mender MCU Certificate for zephyr
+ *
+ * Copyright Northern.tech AS
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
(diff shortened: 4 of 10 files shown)


src/core/client.c
```diff
@@ -329,6 +333,15 @@ mender_client_init(mender_client_config_t *config, mender_client_callbacks_t *ca
         goto END;
     }
 
+#ifdef CONFIG_MENDER_SERVER_DORMANT_CERTIFICATE
+    int rc = mender_add_dormant_cert();
+    if (0 != rc && -EEXIST != rc) {
```
vpodzime, line 338: So `-EEXIST` means the dormant certificate was successfully added? That looks fishy. (also, please use parentheses in the complex expression)

src/core/certs.c
```diff
@@ -0,0 +1,38 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "certs.h"
+#include "log.h"
+#include <errno.h>
+#include <zephyr/net/tls_credentials.h>
```
vpodzime, line 23: We should not have Zephyr-specific code under `src/core`, please put it somewhere under `src/platform`.

vpodzime: And we need at least a weak implementation for this and a no-op POSIX implementation.

vpodzime (commented):
Thanks for working on this!

src/core/client.c
danielskinstad: No, it means that it already exists in the device. I didn't see any reason to throw an error if the certificate was already added; perhaps we should? I can at least make it more explicit if you think we shouldn't

src/core/certs.c
danielskinstad: > We should not have Zephyr-specific code under src/core, please put it somewhere under src/platform.
Oops, you're right

src/include/certs.h
```diff
@@ -0,0 +1,38 @@
...
+#define __MENDER_CERTS_H__
+
+#ifdef __cplusplus
+extern "C" {
+#endif /* __cplusplus */
+
+/**
+ * @brief Add dormant certificate for potential disaster recovery scenarios
+ * @return return 0 on success, -EACCES, -ENOMEM or -EEXIST on error
+ * @note See https://docs.zephyrproject.org/3.7.0/doxygen/html/group__tls__credentials.html#ga640ff6dd3eb4d5017feaab6fab2bb2f7
```
lluiscampos, line 30: Please link to 4.0.0 docs or even latest

src/core/client.c
```diff
@@ -329,6 +333,15 @@ mender_client_init(mender_client_config_t *config, mender_client_callbacks_t *ca
...
+        mender_log_error("Failed to add dormant certificate: %d", rc);
```
lluiscampos, line 339: Should the client fail here or continue? :thinking:

target/zephyr/Kconfig
```diff
@@ -421,6 +431,13 @@ if MENDER_MCU_CLIENT
                 help
                     A secondary security tag that ROOT CA server credential will be referenced with, typically used to authenticate the server from where to download Mender Artifacts. See tls_credential_add.
 
+            config MENDER_NET_CA_CERTIFICATE_TAG_DORMANT
+                int "Dormant CA certificate tag for Server rescue"
+                default 3
```
lluiscampos, line 436: For rotating credentials at some point, likely the user needs for certificates: old server + artifact store + new server + artifact store. Should we set then the TAG here to 9 or something like that? It is just an arbitrary number.

Alternatively we could have like a base number for MENDER_NET_CA_CERTIFICATE_TAG_PRIMARY, and the rest be +1, +2, etc :thinking:

lluiscampos (commented):
I think this feature belongs to mender-mcu :+1:

src/core/client.c
danielskinstad: Unsure. What do you think?

target/zephyr/Kconfig
vpodzime: > Alternatively we could have like a base number for MENDER_NET_CA_CERTIFICATE_TAG_PRIMARY, and the rest be +1, +2, etc 🤔

I like this. It's similar to the `sector offset` I did in the partitioning changes. It gives a user an easy way to leave some room for other stuff if that's what they need.

src/core/client.c
```diff
@@ -329,6 +333,14 @@ mender_client_init(mender_client_config_t *config, mender_client_callbacks_t *ca
         goto END;
     }
 
+#ifdef CONFIG_MENDER_SERVER_DORMANT_CERTIFICATE
+    if (0 != mender_add_dormant_cert()) {
+        /* error logged in function */
+    } else {
+        mender_log_debug("Added dormant certificate");
+    }
```
vpodzime, line 341: You could do
```
if (0 == ...) {
  mender_log_debug(...);
}
/* else error logged in the function */
```
but I don't have a strong preference.

src/include/certs.h
```diff
@@ -0,0 +1,37 @@
...
+
+#ifdef __cplusplus
+extern "C" {
+#endif /* __cplusplus */
+
+/**
+ * @brief Add dormant certificate for potential disaster recovery scenarios
+ * @return return 0 on success, -EACCES, -ENOMEM or -EEXIST on error
+ */
+int mender_add_dormant_cert(void);
```
vpodzime, line 31: Hmm, now that we don't care about the exact `errno`, should we use `mender_err_t` as the return type?

vpodzime (commented):
Looks good to me otherwise.

src/include/certs.h
danielskinstad: Good idea, I'll do that :+1:

src/core/client.c
lluiscampos: What do I think?

TLDR; fatal error :ok_hand: 

As a user,

Similar to what we have for Linux there would be an "integration checklist" for Mender MCU, where I have instructions to make some experiments before I send my device out in the field. There, there will be some special build options to verify that the dormant certificate can be added but also that can actually communicate. So, when building in such mode the error here should be fatal but when doing regular build it should not.

As an engineer,

The above can be a follow-up epic on itself. But in the context of this task the safest is to treat it as a fatal error. If an specific device (read: an specific MbedTLS configuration) has issues with the cert is better to fail hard and let the user fix or remove the feature from Mender MCU Kconfig before shipping devices into the field.

src/core/client.c
```diff
@@ -329,6 +333,14 @@ mender_client_init(mender_client_config_t *config, mender_client_callbacks_t *ca
         goto END;
     }
 
+#ifdef CONFIG_MENDER_SERVER_DORMANT_CERTIFICATE
+    if (MENDER_OK == mender_add_dormant_cert()) {
+        mender_log_debug("Added dormant certificate");
+    } else {
+        /* error logged in function */
```
lluiscampos, line 340: Treat this as fatal error. See above for my reasoning.

lluiscampos (changes requested):

lluiscampos (approved):

cmake/CMake_posix_defaults.txt
```diff
@@ -21,6 +21,7 @@ set(CONFIG_MENDER_PLATFORM_SCHEDULER_TYPE "posix" CACHE STRING "Scheduler type")
 set(CONFIG_MENDER_PLATFORM_STORAGE_TYPE "posix" CACHE STRING "Storage type")
 set(CONFIG_MENDER_PLATFORM_TLS_TYPE "generic/mbedtls" CACHE STRING "TLS type")
 set(CONFIG_MENDER_PLATFORM_SHA_TYPE "generic/mbedtls" CACHE STRING "SHA type")
+set(CONFIG_MENDER_PLATFORM_SHA_TYPE "posix" CACHE STRING "Certificate type")
```
vpodzime, line 24: `CERT_TYPE`

src/platform/certs/zephyr/certs.c
```diff
@@ -0,0 +1,40 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "certs.h"
+#include "log.h"
+#include <errno.h>
+#include <zephyr/net/tls_credentials.h>
```
vpodzime, line 23: We should `#include` third-party headers before our own headers. But not a biggie.

vpodzime (approved):
Just one typo + a nitpick. Thanks!

PR #193, opened 2025-04-04
docs: document flash wear

danielskinstad:
Need some extra  :eyes:  on this
I forgot to take the deployment logs into account on my first write, but I think it's covered. Very open for suggestions if you have some

### DIFF
--- a/docs/flashwear.md
+++ b/docs/flashwear.md
@@ -1,0 +1,45 @@
+## Flash wear estimation
+
+- $N$: Number of deployments during the lifetime of a device.
+- $F$: Probability of failure during a deployment.
+- $W$: Maximum number of state transitions, defined by `MENDER_MAX_STATE_DATA_STORE_COUNT`.
+- $B$: Number of writes for a successful deployment.
+- $K$: Total number of writes to the flash.
+
+We write to the flash in two places:
+1. if no authentication keys are provided by the user, the client will generate authentication keys and write them to the store
+2. during a deployment to keep track of the deployment data, the Artifact being installed and the deployment logs
+
+Most of the flash wear will occur during a deployment. The deployment data, which -- amongst others -- contains the current state during
+a deployment, is written between each state transition. This is what allows Mender to be robust and do atomic deployments. Unless
+`MENDER_DEPLOYMENT_LOGS` is disabled, we write the deployment logs to the flash; these are error or warning logs that have occurred during
+a deployment and that are sent to the Mender server on failed deployments for further diagnostics and troubleshooting.
+
+On a successful deployment we update the local provides database with the meta-data from the Artifact that was installed, combining
+provides, Artifact name and clears provides.
+
+A successful deployment using the zephyr-image update module will go through 6 state transitions and write the provides and the Artifact
+name to the flash. This results in around $K = 6+2$ writes.
+
+Based on numbers seen in production, we can expect a 5% failure rate for deployments. If we factor in the estimated failure percentage, we
+can roughly estimate the expected amount of writes during the lifetime of a device:
+
+$K = N \times ( (W \times F ) + (B \times (1 - F)))$
+
+If we plot in the following values:
+- $N = 100$
+- $F = 0.05$
+- $W = 28$ (the default value of `MENDER_MAX_STATE_DATA_STORE_COUNT`)
+- $B = 8$ (the amount of writes for a successful deployment with zephyr-image update module)
+
+we get an estimated $K = 900$ writes to the flash during the lifetime of a device.
+
+The equation aims at creating an upper estimate; estimating the exact number of writes during a deployment is next to impossible due to the
+unpredictability of the errors and warnings that may occur. However, we can still be fairly sure that the number of writes will be lower than
+the estimate, as its based on the maximum number of state transitions during a deployment. The equation does _not_ take the deployment logs
+directly into account, but considering the low probability of any failed deployment reaching the maximum state count, we can be rather confident
+in the fact that the total number of writes will fall within the estimate.
+
+The flash wear, including the number of erased sectors, physical sectors used, wear leveling, and more, is influenced by the low-level
+implementation of the API, such as NVS in Zephyr. The estimate we provide reflects expected writes, but actual wear is affected by various
+device-specific factors beyond our control.


docs/flashwear.md
```diff
@@ -0,0 +1,39 @@
+## Flash wear estimation
+
+- $N$: Number of deployments during the lifetime of a device.
+- $F$: Probability of failure during a deployment.
+- $W$: Maximum number of state transitions, defined by `MENDER_MAX_STATE_DATA_STORE_COUNT`.
+- $B$: Number of writes for a successful deployment.
+- $K$: Total number of writes to the flash.
+
+We write to the flash in two places:
+1. during initialization if there are no authentication keys in the store
```
lluiscampos, line 10: We need to mention that this is an opt-in feature. Maybe:

_during first boot if the user did not provide a callback to get the keys, Mender will autogenerate one and write it into the store_

docs/flashwear.md
```diff
@@ -0,0 +1,39 @@
...
+2. during a deployment to keep track of the deployment data, the artifact being installed and the deployment logs
+
+Most of the flash wear will occur during a deployment. The deployment data, which -- amongst others -- contains the current state during
+a deployment, is written between each state transition. We have to this to be able to keep track of the deployment. When `MENDER_DEPLOYMENT_LOGS`
```
lluiscampos, line 14:
```suggestion
a deployment, is written between each state transition. This is what allows Mender to be robust and do atomic updates. When `MENDER_DEPLOYMENT_LOGS`
```

docs/flashwear.md
```diff
@@ -0,0 +1,39 @@
...
+is enabled, we also write the deployment logs to the flash; these are error or warning logs that have occured during a deployment.
```
lluiscampos, line 15:
```suggestion
a deployment, is written between each state transition. We have to this to be able to keep track of the deployment. Unless `MENDER_DEPLOYMENT_LOGS`
is disabled, we also write the deployment logs to the flash; these are error or warning logs that have occured during a deployment.
```

:point_up: opt-out feature

lluiscampos, line 15:
```suggestion
is enabled, we also write the deployment logs to the flash; these are error or warning logs that have occurred during a deployment.
```

docs/flashwear.md
```diff
@@ -0,0 +1,39 @@
...
+
+On a successful deployment we store the provides of the deployment and the artifact name of the artifact that was installed.
```
lluiscampos, line 17:
```suggestion
On a successful deployment we update local provides database with the meta-data from the Artifact that was installed, combining provides, Artifact name and clears provides.
```

lluiscampos (changes requested):
The actual formula looks good to me. But here it would be very valuable to have input from @estenberg and @TheYoctoJester 

I did comment in a couple of wording improvements.

lluiscampos:
And thank you @danielskinstad for tackling this one!

TheYoctoJester:
Generally looks good to me, but I'm uncertain about the real value of the entry. While this does a nice job on estimating the number of writing actions to the flash, two things are completely missing:
- an analysis on which sectors are actually written, like 900 times the same one? Or 900 different ones, each time?
- write amplification, as you can't write just one byte in the middle of a sector, you always have to load-modify-store the whole sector.

So I'd like to understand where this is meant to fit in and whom we expect to read and benefit from it.

vpodzime:
> * an analysis on which sectors are actually written, like 900 times the same one? Or 900 different ones, each time?
> 
> * write amplification, as you can't write just one byte in the middle of a sector, you always have to load-modify-store the whole sector.

Both of these are out of our control, though. We use the Zephyr's NVS and FCB implementations which manipulate the sectors themselves. Both, for example, try to avoid erasing sectors so instead of erasing and writing new value, they just write the new value to a new place and then remember which is the latest one. I'm not sure how realistic it would be to deeply analyze those things and keep things up to date.

vpodzime:
I'll look at this one tomorrow. :yawning_face:

estenberg:
I think the information / description here is valuable. For example, when writes happens at all, and approximate number of writes in a (successful) deployment - the normal cases.

I don't think it's worth it to get into the depth of this further. Just an approximate range of number of writes per deployment, and maybe per year assuming e.g. 4 deployments per year would be good enough.

Then mention how to reduce the writes if needed (e.g. deploy less frequently, polling doesn't seem to have an impact).

docs/flashwear.md
```diff
@@ -0,0 +1,41 @@
...
+- $W = 28$ (the default value of `MENDER_MAX_STATE_DATA_STORE_COUNT`)
+- $B = 8$ (the amount of writes for a successful deployment with zephyr-image update module)
+
+we get an estimated $K = 900$ writes to the flash during the lifetime of a device.
+
+The equation aims at creating an upper estimate; estimating the exact number of writes during a deployment is next to impossible due to the
+unpredictability of the errors and warnings that may occur. However, we can still be fairly sure that the number of writes will be lower than
+the estimate, as its based on the maximum number of state transitions during a deployment. The equation does _not_ take the deployment logs
+directly into account, but considering the low probability of any failed deployment reaching the maximum state count, we can be rather confident
+in the fact that the total number of writes will fall within the estimate.
```
vpodzime, line 41: I think it would be nice to add a final note based on Josef's point -- that the number of erased sectors, which physical sectors are used, wear-leveling, etc. depend on the low-level implementation of the API we rely on (e.g. NVS on Zephyr).

vpodzime (approved):
Definitely a good starting point. We shall see how this evolves, but so far it's clear and simple and reads well. :clap:

TheYoctoJester:
> I think the information / description here is valuable. For example, when writes happens at all, and approximate number of writes in a (successful) deployment - the normal cases.

My wording might have been unintentionally harsh @danielskinstad, apologies.
So yeah it definitely provides an estimate of operations that will happen on the flash, and it's good to have that. What I mean to convey is that we need to be clear that is about the operations, and the actual wear is affected by a number of factors beyond our control, respectively which are device specific.

danielskinstad:
> My wording might have been unintentionally harsh @danielskinstad, apologies.

None taken :) I included a note about that at the end

docs/flashwear.md
danielskinstad: Should be done now

lluiscampos (approved):

PR #196, opened 2025-04-07
chore: don't log error if provides is not found

danielskinstad:
This is to avoid logging `mender: Failed to get provides from storage` on a fresh device

### DIFF
--- a/src/core/inventory.c
+++ b/src/core/inventory.c
@@ -270,6 +270,7 @@ mender_inventory_reset_persistent(void) {
 
 static mender_err_t
 collect_persistent_inventory(void) {
+    mender_err_t ret;
     for (uint8_t idx = 0; idx < n_callbacks; idx++) {
         if (callbacks[idx].persistent) {
             keystores_item_t *item = mender_calloc(1, sizeof(keystores_item_t));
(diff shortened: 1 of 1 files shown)


src/core/inventory.c
```diff
@@ -419,7 +419,7 @@ provides_cb(mender_keystore_t **inventory, uint8_t *inventory_len) {
        structure for manipulation with provides data in other places), but we
        need a key-value array so we need to do a transformation. */
     if (MENDER_OK != (ret = mender_storage_get_provides(&provides))) {
-        mender_log_error("Failed to get provides from storage");
+        /* error logged in function */
```
lluiscampos, line 422: We should only hide the error when the error is `MENDER_NOT_FOUND`.

danielskinstad: That was how I initially did it, but then I saw that we log `Unable to read provides` in the function; is there any value in logging `Failed to get provides from storage` as well?

lluiscampos: Right, then I would expect `Unable to read provides` to be printed as a warning or even info message. It is a totally valid case not to have provides, a device can live and get updates without provides and depends functionality from the Artifacts.

I think this is what you addressed now on your second commit. I'll have a look :eyes:

src/core/inventory.c
```diff
@@ -267,15 +267,18 @@ mender_inventory_reset_persistent(void) {
...
             if (NULL == item) {
                 mender_log_error("Unable to allocate memory");
                 return MENDER_FAIL;
             }
-            if (MENDER_OK != callbacks[idx].callback(&(item->keystore), &(item->keystore_len))) {
-                mender_log_error("Failed to get persistent inventory data");
+            if (MENDER_OK != (ret = callbacks[idx].callback(&(item->keystore), &(item->keystore_len)))) {
+                if (MENDER_NOT_FOUND != ret) {
+                    mender_log_error("Failed to get persistent inventory data");
+                }
```
lluiscampos, line 281: I would add an else mender_log_debug here.

PR #197, opened 2025-04-07
MEN-7690: Replace project versioning from file to Git based

lluiscampos:


target/posix/CMakeLists.txt
```diff
@@ -129,5 +129,6 @@ if (CONFIG_MENDER_PLATFORM_TLS_TYPE STREQUAL "mbedtls")
 endif()
 
 # Define version
-file(STRINGS ${MENDER_MCU_VERSION_FILE} MENDER_CLIENT_VERSION)
-add_definitions("-DMENDER_CLIENT_VERSION=\"${MENDER_CLIENT_VERSION}\"")
+include("${MENDER_MCU_ROOT}/cmake/git.cmake")
```
vpodzime, line 132: I'd actually name the file differently because it only covers `git describe`. But not a biggie.

target/zephyr/CMakeLists.txt
```diff
@@ -25,8 +25,12 @@ if(CONFIG_MENDER_MCU_CLIENT)
     zephyr_library_include_directories(${MENDER_PRIV_INCLUDE})
     zephyr_include_directories(${MENDER_MCU_INCLUDE})
 
-    file (STRINGS ${MENDER_MCU_VERSION_FILE} MENDER_CLIENT_VERSION)
-    zephyr_library_compile_definitions(-DMENDER_CLIENT_VERSION=\"${MENDER_CLIENT_VERSION}\")
+    # Include module from Zephyr CMake package. See:
+    # https://github.com/zephyrproject-rtos/zephyr/blob/v4.0.0/cmake/modules/git.cmake
+    include(git)
```
vpodzime, line 30: Here I'd `include()` our file since we have it too. Who knows how they rename or change the file in the future.

vpodzime (commented):
Nice stuff, generally! :+1:

target/posix/CMakeLists.txt
vpodzime: Yeah, but then I'd vote for `git_helpers.cmake` because `git.cmake` looks to me like the base one that `find_package(Git)` should find and load. But YMMV.

vpodzime (approved):
Looks good to me either way.

vpodzime:
Thanks!

PR #200, opened 2025-04-10
MEN-8292: Discard the board qualifier(s) for the default Mender Device Type

lluiscampos:
Using only the base board identifier reads better in the Mender UI, Artifacts, device logs, etc and it will work fine for most of the cases.

For advance use cases, we expect the user to fine tune the device type to his/her needs.


danielskinstad (approved):

vpodzime (approved):
ACK

PR #207, opened 2025-05-09
MEN-8351: fix(build): Set default value for secondary CA cert to false + misc

lluiscampos:
The secondary CA cert is really optional, as a custom Mender Server could use the same domain for both API calls and Artifacts storage.

Set `MENDER_NET_CA_CERTIFICATE_TAG_SECONDARY_ENABLED` default to `n` and instead select it when selecting hosted Mender option(s).

### DIFF
--- a/target/zephyr/Kconfig
+++ b/target/zephyr/Kconfig
@@ -452,7 +452,7 @@ if MENDER_MCU_CLIENT
 
             config MENDER_NET_CA_CERTIFICATE_TAG_DORMANT
                 int "Dormant CA certificate tag for Server rescue"
-                # the tag is arbirtrarily set to 9
+                # the tag is arbitrarily set to 9
                 default 9
                 depends on MENDER_SERVER_DORMANT_CERTIFICATE
                 help
--- a/target/zephyr/Kconfig
+++ b/target/zephyr/Kconfig
@@ -43,10 +43,12 @@ if MENDER_MCU_CLIENT
 
             config MENDER_SERVER_HOST_US
                 bool "https://hosted.mender.io"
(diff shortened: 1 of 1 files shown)


target/zephyr/Kconfig
```diff
@@ -469,7 +469,7 @@ if MENDER_MCU_CLIENT
                 range 0 2
                 default 2
                 help
-                    Peer verification level for TLS connection.
+                    Peer verification level for TLS connection. For testing purposes, set to 0 to skip TLS peer verify
```
vpodzime, line 474: So what does `1` mean?

vpodzime (approved):
Looks good to me otherwise.

danielskinstad (approved):
Looks good to me

target/zephyr/Kconfig
lluiscampos: It means optional. See [here](https://github.com/Mbed-TLS/mbedtls/blob/1782587af27c693e6f0faf5851d76c75257bec08/include/mbedtls/ssl.h#L329) and [here](https://github.com/Mbed-TLS/mbedtls/blob/1782587af27c693e6f0faf5851d76c75257bec08/include/mbedtls/ssl.h#L2035).

lluiscampos: IMO we don't really support "optional", because to properly use it we would need to check the handshake result and print a warning or something.

vpodzime: Right. The `help` should say it then. And perhaps create a task to make sure `1` is handled gracefully?

lluiscampos: > The help should say it then.

Amended

> And perhaps create a task to make sure 1 is handled gracefully?

[Done](https://northerntech.atlassian.net/browse/MEN-8413).

PR #213, opened 2025-08-19
[WIP] Zephyr 4.2.0

danielskinstad:


src/platform/net/zephyr/http.c
```diff
@@ -370,10 +385,20 @@ http_response_cb(struct http_response *response, MENDER_ARG_UNUSED enum http_fin
             mender_log_error("An error occurred, stop reading data");
         }
     }
+#if ZEPHYR_VERSION_BELOW_FOUR_TWO
+    return;
+#else
+    return 0;
+#endif /* ZEPHYR_VERSION_BELOW_FOUR_TWO */
```
vpodzime, line 392: We could do
```
#if ZEPHYR_VERSION_BELOW_FOUR_TWO
#define HTTP_CALLBACK_RETURN_VALUE
#else
#define HTTP_CALLBACK_RETURN_VALUE 0
#endif
```
and use the latter macro without and `#if-#else`. But not a big difference...

vpodzime: The same applies to `HTTP_CALLBACK_RETURN_TYPE`. And while I think about the two macros, I prefer them over having `#if-#else` in multiple places. Keep that contained in one place where we check the Zephyr version macros and define our own based on it.

lluiscampos (changes requested):
There are two changes in this PR, one is backwards compatible and the other one not.

I think that we should find a way to introduce `MCUBOOT_MODE_SWAP_USING_MOVE` conditionally or just remove the conditional logic for the `HTTP_CALLBACK_RETURN_...` and explicitly mark it as **Zephyr 4.2 only**. I would be okey with the second option (keep it explicit, as we are only going to test one version)

On that note, the [README](https://github.com/mendersoftware/mender-mcu/pull/213/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R54) should also be updated regardless.

Requesting changes but open for which direction to go 🍿

lluiscampos:
> Requesting changes but open for which direction to go 🍿

After face-to-face discussion (and reading the JIRA ticket) we agreed in keeping "in the code" the compatibility and leave the KConfig part as-is (we know this is not really in use when building with SysBuild).

So my request changes are:
* Update the README
* Mark at least one of the changes as `feat` (iow support new Zephyr release is a feature in itself). While being in preview it is not very important but it is a good practice I think.

vpodzime (approved):
:heart:

PR #229, opened 2026-04-07
fix: Add explicit 413 handling when publishing deployment logs

danielskinstad:
Return a regular failure when encountering 413 Content too large errors when publishing deployment logs. This is done to ensure we don't retry unnecessarily.

Ticket: MEN-9389


vpodzime (approved):
Looks good to me. I just realized, though, that this is nearly impossible to happen because the deployment logs are stored in a circular buffer/flash so they will never grow that big. Unless someone has a really big piece of flash to waste and configures things in such a way...

PR #230, opened 2026-04-14
feat: Add support for chunked artifact download

Dexter9532:
Related to https://hub.mender.io/t/mender-mcu-integration-on-nrf9151dk-http-request-fails-during-artifact-download/8253/4

### DIFF
--- a/src/platform/net/zephyr/http.c
+++ b/src/platform/net/zephyr/http.c
@@ -57,6 +57,15 @@
  * @brief Rate limit HTTP status code
  */
 #define MENDER_HTTP_RATE_LIMIT_HTTP_CODE 429
+#if defined(CONFIG_MENDER_ARTIFACT_DOWNLOAD_RANGES)
+#define MENDER_HTTP_CHUNK_RETRY_MAX      CONFIG_MENDER_ARTIFACT_DOWNLOAD_CHUNK_MAX_TRIES
+#define MENDER_HTTP_CHUNK_RETRY_DELAY_MS CONFIG_MENDER_ARTIFACT_DOWNLOAD_CHUNK_DELAY_MS
+#define MENDER_HTTP_ARTIFACT_RANGE_MAX   CONFIG_MENDER_ARTIFACT_DOWNLOAD_CHUNK_SIZE
+#else
+#define MENDER_HTTP_CHUNK_RETRY_MAX      1
+#define MENDER_HTTP_CHUNK_RETRY_DELAY_MS 0
+#define MENDER_HTTP_ARTIFACT_RANGE_MAX   0
+#endif
 
 const size_t mender_http_recv_buf_length = 512;
 
@@ -446,16 +455,22 @@ mender_http_artifact_download(const char *uri, mender_artifact_download_data_t *
     mender_err_t        ret                = MENDER_FAIL;
     struct http_request request            = { 0 };
     mender_err_t        request_ret        = MENDER_OK;
-    const char         *header_fields[3]   = { NULL }; /* The list is NULL terminated; make sure the size reflects it */
+    const char         *header_fields[4]   = { NULL }; /* The list is NULL terminated; make sure the size reflects it */
     size_t              header_fields_size = sizeof(header_fields) / sizeof(header_fields[0]);
     char               *host               = NULL;
     char               *port               = NULL;
     char               *url                = NULL;
     int                 sock               = -1;
     int                 http_req_ret;
+    bool                connected  = false;
+    bool                done       = false;
+    size_t              offset     = 0;
+    size_t              total_size = 0;
+    const bool          use_ranges = IS_ENABLED(CONFIG_MENDER_ARTIFACT_DOWNLOAD_RANGES);
 
     /* Headers to be added to the request */
-    char *host_header = NULL;
+    char *host_header  = NULL;
+    char *range_header = NULL;
 
     /* Retrieve host, port and url */
     if (MENDER_OK != mender_net_get_host_port_url(uri, http_config.host, &host, &port, &url)) {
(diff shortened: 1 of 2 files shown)


vpodzime:
Thanks for bringing this, @Dexter9532! It might be a very nice contribution for many other boards, incl. our reference board where this could potentially allow us to shrink some buffers.

Instead of going directly to the code changes, let me suggest one thing -- I think this should not be a fix for a particular board, but rather a new feature of the Mender MCU client. Which means that there should be new KConfig options for enabling ranged artifact downloads and specifying the parameters (max chunk size, max retries,...). The defaults for those can be based on board selection/configuration, of course.

Dexter9532:
Hello @vpodzime,
Sounds good! Would you like me to iterate on this pr, or how do you recommend we move on?

Dexter9532:
Here would be the way I would do it. What are your thoughts?

My plan:
- Add three Kconfig options to target/zephyr/Kconfig:
  - CONFIG_MENDER_ARTIFACT_DOWNLOAD_RANGES (bool)
  - CONFIG_MENDER_ARTIFACT_DOWNLOAD_CHUNK_SIZE (int)
  - CONFIG_MENDER_ARTIFACT_DOWNLOAD_MAX_RETRIES (int)

- Add to /modules/mender-mcu/src/platform/net/zephyr/http.c:

if (IS_ENABLED(CONFIG_MENDER_ARTIFACT_DOWNLOAD_RANGES)) {
    /* send Range header / retry logic */
} else {
    /* current behavior */
}

vpodzime:
> Here would be the way I would do it. What are your thoughts?

Sounds like a great plan to me! Please update this PR with such a nice and great version of the changes. :grin:

Dexter9532:
I was a bit unsure whether I should restore the original HTTP request flow and keep the ranged-download handling as a separate path. That would duplicate some code, so for now I kept the shared flow and made the ranged behavior configurable instead.

Do you think it would be better to split those into two separate paths?

target/zephyr/Kconfig
```diff
@@ -557,6 +557,36 @@ if MENDER_MCU_CLIENT
...
+
+            config MENDER_ARTIFACT_DOWNLOAD_RANGES
+                bool "Enable ranged artifact download"
+                default n
+                help
+                    Enable downloading Mender Artifacts in HTTP range-based chunks instead of a single request.
+                    This can help on targets where large downloads are unreliable or where smaller chunks reduce
+                    memory and transport pressure.
+
+            config MENDER_ARTIFACT_DOWNLOAD_CHUNK_SIZE
```
vpodzime, line 569: This and the two other config values should `depends on` on the one above enabling this functionality.

target/zephyr/Kconfig
```diff
@@ -557,6 +557,36 @@ if MENDER_MCU_CLIENT
...
+                int "Artifact download chunk size"
+                default 2048
+                range 1 65536
+                help
+                    Maximum number of bytes requested per HTTP range request.
+
+            config MENDER_ARTIFACT_DOWNLOAD_MAX_TRIES
+                int "Artifact download max retries"
```
vpodzime: Please keep the word `CHUNK/chunk` in the config option names and descriptions (also below) so that people don't confuse them with global retries (with chunks disabled).

vpodzime (commented):
Thanks! ❤️  I'll get to the actual code tomorrow, here is just a first round of suggestions/change requests in the KConfig. I've also updated the PR description.

vpodzime:
> That would duplicate some code, so for now I kept the shared flow and made the ranged behavior configurable instead.

I'd do the same. At the first glance this doesn't look like the code will be much worse due to this so let's stick to this approach.

src/platform/net/zephyr/http.c
```diff
@@ -463,77 +470,145 @@ mender_http_artifact_download(const char *uri, mender_artifact_download_data_t *
...
-
-    /* Connect to the server */
-    sock = mender_net_connect(host, port);
-    if (sock < 0) {
-        mender_log_error("Unable to open HTTP client connection");
-        goto END;
-    }
     if (MENDER_OK != (ret = dl_data->artifact_download_callback(MENDER_HTTP_EVENT_CONNECTED, NULL, 0, dl_data))) {
         mender_log_error("An error occurred while calling 'MENDER_HTTP_EVENT_CONNECTED' artifact callback");
         goto END;
```
vpodzime: Here we call the callback before anything is connected.

src/platform/net/zephyr/http.c
```diff
@@ -463,77 +470,145 @@ mender_http_artifact_download(const char *uri, mender_artifact_download_data_t *
...
     }
+    connected = true;
```
vpodzime: This variable is only unconditionally assigned `true` here and `false` below, it doesn't serve any purpose.

src/platform/net/zephyr/http.c
```diff
@@ -463,77 +470,145 @@ mender_http_artifact_download(const char *uri, mender_artifact_download_data_t *
...
+
+    while (true) {
```
vpodzime: I'd prefer a `do-while` loop here and it should be `while (!done)` with `done` being assigned in the loop based on a clear condition.

src/platform/net/zephyr/http.c
```diff
@@ -463,77 +470,145 @@ mender_http_artifact_download(const char *uri, mender_artifact_download_data_t *
...
+            mender_log_error("An error occurred, connection has been closed");
+            dl_data->artifact_download_callback(MENDER_HTTP_EVENT_ERROR, NULL, 0, dl_data);
+            mender_free(request.recv_buf);
+            goto END;
+        }
+        *status             = request.internal.response.http_status_code;
+        body_bytes_received = request.internal.response.processed;
+
+        mender_free(host_header);
+        host_header = NULL;
```
vpodzime: We have `FREE_AND_NULL()`.

vpodzime (commented):
A couple more suggestions in the code.

src/platform/net/zephyr/http.c
Dexter9532: If we remove this flag, wouldn’t `CONNECTED` be emitted on every chunk/reconnect during ranged downloads? Should we keep the guard and just rename the variable to better reflect its purpose?

vpodzime: Now it plays its role in the code. :+1: But it's still unconditionally set to `false` below the `do-while` block and then part of a condition (where it will thus always evaluate to the same).

Dexter9532: Yes, nice !

Dexter9532:
@vpodzime Great suggestions/change requests! :smiley:

src/platform/net/zephyr/http.c
```diff
@@ -446,13 +449,18 @@ mender_http_artifact_download(const char *uri, mender_artifact_download_data_t *
...
     char               *host               = NULL;
     char               *port               = NULL;
     char               *url                = NULL;
     int                 sock               = -1;
     int                 http_req_ret;
+    bool                connected  = false;
+    bool                done       = false;
+    size_t              offset     = 0;
+    size_t              total_size = 0;
+    bool                use_ranges = false;
```
vpodzime: `use_ranges` can be initialized with the proper value here right away. And it should be `const`.

src/platform/net/zephyr/http.c
```diff
@@ -463,77 +471,143 @@ mender_http_artifact_download(const char *uri, mender_artifact_download_data_t *
...
+                    if (MENDER_OK != (ret = dl_data->artifact_download_callback(MENDER_HTTP_EVENT_CONNECTED, NULL, 0, dl_data))) {
+                        mender_log_error("An error occurred while calling 'MENDER_HTTP_EVENT_CONNECTED' artifact callback");
+                        mender_net_disconnect(sock);
+                        sock = -1;
+                        goto END;
+                    }
+                    connected = true;
+                }
+                http_req_ret = http_client_req(sock, &request, MENDER_HTTP_REQUEST_TIMEOUT, dl_data);
+                mender_net_disconnect(sock);
```
vpodzime: This unconditionally disconnects, which means every single request creates a new connection. That's an expensive operation (esp. with HTTPS/TLS). It might be necessary for resetting a failed state for the following attempts, but if things are OK and the connection works, all requests (for all chunks) should be performed without reconnecting.

src/platform/net/zephyr/http.c
```diff
@@ -463,77 +471,143 @@ mender_http_artifact_download(const char *uri, mender_artifact_download_data_t *
...
+    }
+
     ret = MENDER_OK;
 
 END:
-
-    /* Close connection */
-    if (sock >= 0) {
-        mender_net_disconnect(sock);
+    if ((MENDER_OK != ret) && connected) {
```
vpodzime: Why is `connected` part of the condition here?

vpodzime (commented):
Looks generally good to me otherwise. Thanks for adjusting this! 👏

src/platform/net/zephyr/http.c
Dexter9532: connected is just there to make the callback flow sensible. We only set it after sending MENDER_HTTP_EVENT_CONNECTED, so the cleanup path only sends MENDER_HTTP_EVENT_ERROR if the session was actually established from the callback’s point of view.

Without that check, we could end up sending ERROR for failures that happened before CONNECTED, which would make the event sequence a bit odd.

vpodzime: Well, I'd say that failing to connect should still result in `MENDER_HTTP_EVENT_ERROR`.

Dexter9532: Yes of course, will change it!

src/platform/net/zephyr/http.c
```diff
@@ -463,75 +471,144 @@ mender_http_artifact_download(const char *uri, mender_artifact_download_data_t *
...
+                    connected = true;
+                }
+                http_req_ret = http_client_req(sock, &request, MENDER_HTTP_REQUEST_TIMEOUT, dl_data);
+                if (http_req_ret >= 0) {
+                    break;
+                }
+                mender_net_disconnect(sock);
+                sock = -1;
+                mender_log_warning(
+                    "HTTP request failed at offset %zu: %s (attempt %d/%d)", offset, strerror(-http_req_ret), attempt, MENDER_HTTP_CHUNK_RETRY_MAX);
```
vpodzime: I'd put the above four lines into an `else` branch, it would make it easier to spot that the disconnect doesn't happen in case of a successful request.

vpodzime (commented):
Just one more nitpick. Looks good to me now otherwise. Could you please retest your changes and make sure they still work in your use case? I'll test it myself tomorrow with our reference board (both chunked and non-chunked download).

Dexter9532:
> Just one more nitpick. Looks good to me now otherwise. Could you please retest your changes and make sure they still work in your use case? I'll test it myself tomorrow with our reference board (both chunked and non-chunked download).

I have created a PR our mender integration repository please see https://github.com/id8-engineering/nrf9151-dk-mender/pull/29 

And I can say that ranged based artifact downloads works:
```
*** Booting MCUboot v2.3.0-dev-0d9411f5dda3 ***
*** Using nRF Connect SDK v3.2.1-d8887f6f32df ***
*** Using Zephyr OS v4.2.99-ec78104f1569 ***
I: Starting bootloader
I: Primary image: magic=unset, swap_type=0x1, copy_done=0x3, image_ok=0x3
I: Secondary image: magic=unset, swap_type=0x1, copy_done=0x3, image_ok=0x3
I: Boot source: none
I: Image index: 0, Swap type: none
I: Bootloader chainload address offset: 0x10000
I: Image version: v0.0.0
�*** Booting nRF Connect SDK v3.2.1-d8887f6f32df ***
*** Using Zephyr OS v4.2.99-ec78104f1569 ***
[00:00:00.255,676] <inf> main: Starting LTE
[00:00:19.931,732] <inf> main: LTE connected
[00:00:19.931,762] <inf> mender: Device type: [nrf9151-dk-mender]
[00:00:19.938,018] <inf> main: Mender activated
[00:00:19.938,018] <inf> main:  V0.1
[00:00:19.938,262] <inf> mender: Initialization done
[00:00:19.938,293] <inf> mender: Checking for deployment...
[00:00:23.789,794] <inf> mender: Downloading artifact with id '7276eaf...', name 'V0.1', uri 'https://c271964d41749feb10da762816c952ee.r'
[00:00:29.272,521] <inf> mender: Start flashing artifact 'zephyr.signed.bin' with size 339474
[00:00:36.491,394] <inf> mender: Downloading 'zephyr-image' 10%... [34304/339474]
[00:00:36.778,411] <wrn> mender: HTTP request failed at offset 43008: Message too long (attempt 1/3)
[00:00:45.819,305] <inf> mender: Downloading 'zephyr-image' 20%... [68096/339474]
[00:00:47.262,908] <wrn> mender: HTTP request failed at offset 81920: Message too long (attempt 1/3)
[00:00:55.474,243] <inf> mender: Downloading 'zephyr-image' 30%... [101888/339474]
[00:00:57.104,370] <wrn> mender: HTTP request failed at offset 116736: Message too long (attempt 1/3)
[00:01:04.955,505] <inf> mender: Downloading 'zephyr-image' 40%... [136192/339474]
[00:01:07.825,866] <wrn> mender: HTTP request failed at offset 157696: Message too long (attempt 1/3)
[00:01:14.624,511] <inf> mender: Downloading 'zephyr-image' 50%... [169984/339474]
[00:01:18.735,382] <wrn> mender: HTTP request failed at offset 196608: Message too long (attempt 1/3)
[00:01:23.915,649] <inf> mender: Downloading 'zephyr-image' 60%... [203776/339474]
[00:01:29.435,699] <wrn> mender: HTTP request failed at offset 237568: Message too long (attempt 1/3)
[00:01:33.338,592] <inf> mender: Downloading 'zephyr-image' 70%... [238080/339474]
[00:01:39.434,326] <wrn> mender: HTTP request failed at offset 274432: Message too long (attempt 1/3)
[00:01:43.216,003] <inf> mender: Downloading 'zephyr-image' 80%... [271872/339474]
[00:01:49.560,668] <inf> mender: Downloading 'zephyr-image' 90%... [305664/339474]
[00:01:49.828,796] <wrn> mender: HTTP request failed at offset 313344: Message too long (attempt 1/3)
[00:01:59.304,260] <inf> mender: Download done, installing artifact
[00:02:00.887,969] <inf> mender: Artifact installation done, rebooting
[00:02:02.413,757] <wrn> main: Mender requested reboot; waiting 5s for logs
*** Booting MCUboot v2.3.0-dev-0d9411f5dda3 ***
*** Using nRF Connect SDK v3.2.1-d8887f6f32df ***
*** Using Zephyr OS v4.2.99-ec78104f1569 ***
I: Starting bootloader
I: Primary image: magic=unset, swap_type=0x1, copy_done=0x3, image_ok=0x3
I: Secondary image: magic=good, swap_type=0x2, copy_done=0x3, image_ok=0x3
I: Boot source: none
I: Image index: 0, Swap type: test
I: Starting swap using move algorithm.
I: Bootloader chainload address offset: 0x10000
I: Image version: v0.0.0
*** Booting nRF Connect SDK v3.2.1-d8887f6f32df ***
*** Using Zephyr OS v4.2.99-ec78104f1569 ***
[00:00:00.255,737] <inf> main: Starting LTE
[00:00:22.641,693] <inf> main: LTE connected
[00:00:22.641,754] <inf> mender: Device type: [nrf9151-dk-mender]
[00:00:22.647,216] <inf> main: Mender activated
[00:00:22.647,247] <inf> main:  V0.1
[00:00:22.648,193] <inf> mender: Initialization done
[00:00:26.318,603] <inf> mender: Application has been mark valid and rollback canceled
```

I was tho a little curious of the warning `[00:00:36.778,411] <wrn> mender: HTTP request failed at offset 43008: Message too long (attempt 1/3)` coming after every 10%, is this only my board or is there something wrong in the code causing it to LOG this warning all the time?

vpodzime:
> is this only my board or is there something wrong in the code causing it to LOG this warning all the time?

I'll see if it happens our our reference board too. But I won't get to it before Monday, unfortunately.

Dexter9532:
> I'll see if it happens our our reference board too. But I won't get to it before Monday, unfortunately.

No worries!

src/platform/net/zephyr/http.c
```diff
@@ -57,6 +57,9 @@
  * @brief Rate limit HTTP status code
  */
 #define MENDER_HTTP_RATE_LIMIT_HTTP_CODE 429
+#define MENDER_HTTP_CHUNK_RETRY_MAX      CONFIG_MENDER_ARTIFACT_DOWNLOAD_CHUNK_MAX_TRIES
+#define MENDER_HTTP_CHUNK_RETRY_DELAY_MS CONFIG_MENDER_ARTIFACT_DOWNLOAD_CHUNK_DELAY_MS
+#define MENDER_HTTP_ARTIFACT_RANGE_MAX   CONFIG_MENDER_ARTIFACT_DOWNLOAD_CHUNK_SIZE
```
vpodzime, line 63: The new `CONFIG_` values are undefined when `CONFIG_MENDER_ARTIFACT_DOWNLOAD_RANGES` is not enabled.

vpodzime (changes requested):
Seems to work great, otherwise! 🥳 👏

src/platform/net/zephyr/http.c
Dexter9532: Yeah, nice! I never tried without it.

vpodzime:
Thanks @Dexter9532! Just one last nitpick -- could you please change the commit message to:
```
feat: Add support for chunked artifact download

Using HTTP requests with `Range` headers artifact
downloads can be chunked reducing requirements
on network buffers and potentially performing better
in bad network deployments.

Signed-off-by: Bobo Bäck Engström <bobo@id8-engineering.io>
```
This was it shall provide a nice entry in the changelogs of the next release.

vpodzime (approved):
Looks great to me now!

Dexter9532:
> Looks great to me now!

Nice, thanks for the help to get this fixed!

src/platform/net/zephyr/http.c
```diff
@@ -463,75 +477,145 @@ mender_http_artifact_download(const char *uri, mender_artifact_download_data_t *
...
+        request.protocol = "HTTP/1.1";
+        request.response = artifact_response_cb;
+        if (NULL == (request.recv_buf = (uint8_t *)mender_malloc(mender_http_recv_buf_length))) {
+            mender_log_error("Unable to allocate memory");
+            goto END;
+        }
+        request.recv_buf_len = mender_http_recv_buf_length;
 
-    /* Add headers */
-    host_header = header_alloc_and_add(header_fields, header_fields_size, "Host: %s\r\n", host);
```
vpodzime, line 479: Actually! A second pair of eyes spotted that this `Host:` header is now not being added/set. That's a regression, it's needed in some setups.

vpodzime (changes requested):

vpodzime (approved):
Still works for me and the code looks good.

Dexter9532:
Nice, thanks so much for this and super cool to have contributed to mender-mcu!

PR #236, opened 2026-04-28
feat: Warnings based on device tier defaults

rewanrashid-boop:
Added defaults based on device tier in Zephyr Kconfig and extended functionality in Cmake source file to trigger warnings.

Changelog: Warnings based on device tier defaults
Ticket: MEN-9288


cmake/mender_mcu_sources.txt
michalkopczan, line 51: Because of the OR, this warning will show even if we have standard tier, and interval > 1800. Or am I mistaken?
What happens if there's no tier set? Does it default to micro?

michalkopczan (commented):
Nice.
Does it solve the thing about Kconfig not being updated when the tier changes (when some config is already written in the config file)?

cmake/mender_mcu_sources.txt
rewanrashid-boop: Basically if there is not tier set in prj.conf so no CONFIG_MENDER_DEVICE_TIER then the tier will be set to micro is what happens here. Now that i look at it though maybe i should change the tier checks for the standard warnings further down to get rid of the NOT CONFIG_MENDER_DEVICE_TIER check and just check if the tier is standard, because we assume it has already been set by the user in prj.conf.

danielskinstad: > Now that i look at it though maybe i should change the tier checks for the standard warnings further down to get rid of the NOT CONFIG_MENDER_DEVICE_TIER check and just check if the tier is standard, because we assume it has already been set by the user in prj.conf.

This seems like the way to go

danielskinstad (commented):
The `NOT CONFIG_MENDER_DEVICE_TIER` is a bit buggy/weird, so I think you should go with your comment. 

I also wouldn't really call this a "feat". "build" would be better IMO.

Did you create a follow-up ticket for the defaults handling?

rewanrashid-boop:
I'll change from feat to build then and remove the `NOT CONFIG_MENDER_DEVICE_TIER`. Didn't create a  follow up ticket for it yet but will do it today (kind of forgot ty for reminder) :D

danielskinstad (approved):
LGTM, thanks

rewanrashid-boop:
Link to task: https://northerntech.atlassian.net/browse/MEN-9288

PR #240, opened 2026-06-03
fix: correctly deref deployment_data double pointer for null check

danielb-IDG:
Null check as it stood will always succeed, and doesn't check the correct level of indirection. Causes an assert in validate_deployment_data() if cJSON_Parse() returns a null


vpodzime (approved):
Thanks!

vpodzime:
@danielb-IDG please add a `Signed-off-by` line to your commit message (e.g. by running `git commit --amend -s`).

danielb-IDG:
Ok, done, Thanks for the quick turnaround!

vpodzime:
> Ok, done, Thanks for the quick turnaround!

Likewise! Thanks for the fix!

PR #246, opened 2026-06-22
feat: make logging definitions generic

aduskett:
In an effort to make porting mender-mcu to other platforms such as FreeRtOS easier, it's important to make APIs generic, easy to overwrite, and with sane defaults. One such as example is logging.

Vanilla FreeRtOS has a printf method, but anyone forced to use the TI FreeRtOS has to use the DebugP_log methods.

Declare the init, exit, print, deployment_logs_activate, and mender_deployment_logs_deactivate methods in a mender_log_t struct so users can integrate their own logging solution easily while linking to mender-mcu.

### DIFF
--- a/include/mender/log.h
+++ b/include/mender/log.h
@@ -46,6 +46,36 @@ extern "C" {
 #define CONFIG_MENDER_LOG_LEVEL MENDER_LOG_LEVEL_INF
 #endif /* CONFIG_MENDER_LOG_LEVEL */
 
+typedef struct {
+    /**
+     * @brief Initialize mender log
+     * @return MENDER_OK if the function succeeds, error code otherwise
+     */
+    mender_err_t (*init)(void);
+
+    /**
+    * @brief Release mender log
+    * @return MENDER_OK if the function succeeds, error code otherwise
+    */
+    mender_err_t (*exit)(void);
+
+    mender_err_t (*print)(uint8_t level, const char *filename, const char *function, int line, char *format, ...);
+#ifdef CONFIG_MENDER_DEPLOYMENT_LOGS
+    /**
+     * @brief Activate deployment logs saving
+     * @return MENDER_OK if the function succeeds, error code otherwise
+     */
+    mender_err_t (*deployment_logs_activate)(void);
+
+    /**
+     * @brief Deactivate deployment logs saving
+     * @return MENDER_OK if the function succeeds, error code otherwise
+     */
+    mender_err_t (*mender_deployment_logs_deactivate)(void);
+#endif /* CONFIG_MENDER_DEPLOYMENT_LOGS */
+} mender_log_t;
+extern mender_log_t mender_log;
+
 #ifdef __ZEPHYR__
 
 LOG_MODULE_DECLARE(mender, CONFIG_MENDER_LOG_LEVEL);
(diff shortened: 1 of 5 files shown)


aduskett:
What exactly failed for the pipeline? And can the mender-mcu gitlab project be made public so I don't have to ask every time? :)

danielskinstad:
> can the mender-mcu gitlab project be made public so I don't have to ask every time?

Looking into it.

```
commit 413645654cd15aa33f518c8d6787ea2f45f06f86
Author: Adam Duskett <aduskett@gmail.com>
Date:   Fri Jun 19 11:55:33 2026 +0200
    feat: make logging definitions to a struct
    
    In an effort to make porting mender-mcu to other platforms such as FreeRtOS
    easier, it's important to make APIs generic, easy to overwrite, and with sane
    defaults. One such as example is logging.
    
    Vanilla FreeRtOS has a printf method, but anyone forced to use the
    TI FreeRtOS has to use the DebugP_log methods.
    
    Declare the init, exit, print, deployment_logs_activate, and
    mender_deployment_logs_deactivate methods in a mender_log_t struct
    so users can integrate their own logging solution easily while
    linking to mender-mcu.
Commit 413645654cd15aa33f518c8d6787ea2f45f06f86 is not signed off! Use --signoff with your commit. Make sure that the Author of the commit matches the one in Signed-off-by
```

vpodzime (commented):
### Some intro/background

This is similar to what I originally wanted to do when making changes for logging "the right way" on Zephyr (see 707145f15c7c461a1d71cbf241a85f93f7500542). However, the issue was that Zephyr logging was based on macros and we ended up in the `#ifdef __ZEPHYR__` in _include/mender/log.h_ that you can see.

### More feedback

What is the benefit of having the struct over just choosing the right definitions/implementations of the respective functions at build time? If you look at the other platform code, that's how it's done -- there is a common header with declarations and then the platform-specific implementation is chosen at build time. And the platform can easily be _esp-idf_, _ti-free-rtos_,..., whatever.

I could see some benefit if a struct like the one added in this PR was part of `mender_client_config_t` and thus passed to `mender_client_init()`. That would allow the user code to pass some application-specific logging functions (perhaps showing the logs from Mender somewhere,...). But again, that would not work on Zephyr, for example, where we need to use the macros.

aduskett:
I guess the idea is that there wouldn't be a need for a separate implementation of each platform (like ti-freertos) if 99% of the code is reusable and the user can simply fill in a struct. I agree that the zephyr logging is a pain point, and I would like to get an x86 qemu session going for it (along with 4.4.1 compatibility)  for further testing.

vpodzime:
> I guess the idea is that there wouldn't be a need for a separate implementation of each platform (like ti-freertos) if 99% of the code is reusable and the user can simply fill in a struct.

What do you think would be the reusable parts?

aduskett:
> > I guess the idea is that there wouldn't be a need for a separate implementation of each platform (like ti-freertos) if 99% of the code is reusable and the user can simply fill in a struct.
> 
> What do you think would be the reusable parts?

Sorry for the late reply! It's difficult to say right now, as we are in the process of porting mender-mcu over to FreeRtOS and probably soon SafeRtOS!

vpodzime:
@aduskett I was about to get back to this. But only to close it, we decided to follow the existing pattern for [ESP-IDF logging](https://github.com/mendersoftware/mender-mcu/pull/251) and I want to actually change that to use macros directly like on Zephyr because that's the only way to give the compiler a chance to respect the log level settings and omit the strings and function calls that would have no effect.

PR #247, opened 2026-06-30
Support Zephyr 4.4.0

danielskinstad:


vpodzime (approved):
Looks OK to me. But the first commit makes me think we should follow-up with [MEN-9221](https://northerntech.atlassian.net/browse/MEN-9221) ASAP. And both commits together make me quite afraid of future similar changes making the code less and less readable if we want to support multiple versions of Zephyr.

michalkopczan:
Due to no empty line between the Changelog and "assisted-by", the changelog includes "Assisted-by: Claude Code:claude-opus-4-8", I think we don't want that.

michalkopczan (approved):
Besides claude in changelogs, LGTM.

PR #251, opened 2026-07-15
ESP-IDF component

danielskinstad:


danielskinstad:
See https://github.com/mendersoftware/mender-mcu-integration/pull/116

src/platform/log/esp-idf/log.c
```diff
@@ -0,0 +1,70 @@
...
+mender_err_t
+mender_log_init(void) {
+    /* Nothing to do */
+    return MENDER_OK;
+}
+
+mender_err_t
+mender_log_print(uint8_t level, const char *filename, const char *function, int line, char *format, ...) {
+    /* Silence -Wunused-parameter warning */
+    (void)filename;
```
vpodzime: we have `ARG_UNUSED` in our repo, I believe

vpodzime (approved):
Looks good to me otherwise.

target/esp-idf/component.cmake
```diff
@@ -0,0 +1,82 @@
+# @file      component.cmake
+# @brief     mender-mcu ESP-IDF component file
+#
+# Copyright Northern.tech AS
```
michalkopczan, line 4: Should we include the year in the copyright for this and other files?

michalkopczan (approved):

src/platform/log/esp-idf/log.c
danielskinstad: Right, thanks

target/esp-idf/component.cmake
danielskinstad: @vpodzime knows best here. We don't do it for the other files in this repo AFAICT

vpodzime: It should be there, but I'd prefer consistency here and fixing it in the whole repo.

PR #254, opened 2026-07-22
FreeRTOS scheduler 

danielskinstad:


src/platform/os/freertos/os.c
```diff
@@ -0,0 +1,425 @@
...
+ */
+#ifndef CONFIG_MENDER_SCHEDULER_WORK_QUEUE_PRIORITY
+#define CONFIG_MENDER_SCHEDULER_WORK_QUEUE_PRIORITY (tskIDLE_PRIORITY + 1)
+#endif /* CONFIG_MENDER_SCHEDULER_WORK_QUEUE_PRIORITY */
+
+/**
+ * @brief Default work queue length
+ */
+#ifndef CONFIG_MENDER_SCHEDULER_WORK_QUEUE_LENGTH
+#define CONFIG_MENDER_SCHEDULER_WORK_QUEUE_LENGTH (10)
```
vpodzime, line 51: Any reason for 10 being a good default here?

src/platform/os/freertos/os.c
```diff
@@ -0,0 +1,425 @@
...
+        mender_log_error("Unable to create work queue");
+        return MENDER_FAIL;
+    }
+
+    /* Create semaphore used to wait for the work queue task termination */
+    mender_os_scheduler_exit_sem_handle = xSemaphoreCreateBinary();
+    if (NULL == mender_os_scheduler_exit_sem_handle) {
+        mender_log_error("Unable to create exit semaphore");
+        vQueueDelete(mender_os_scheduler_work_queue_handle);
+        mender_os_scheduler_work_queue_handle = NULL;
```
vpodzime: We have `DESTROY_AND_NULL()` (applies to a few more places below too).

src/platform/os/freertos/os.c
```diff
@@ -0,0 +1,425 @@
...
+    /* Create work context */
+    mender_platform_work_t *work_context = mender_calloc(1, sizeof(mender_platform_work_t));
+    if (NULL == work_context) {
+        mender_log_error("Unable to allocate memory");
+        goto FAIL;
+    }
+
+    /* Copy work parameters.
+     * Note: backoff parameters are ignored; failed work simply retries at its
+     * regular period. Proper backoff and rate-limit handling is a follow-up. */
```
vpodzime, line 144: Let's make sure we take notes of things like this.

src/platform/os/freertos/os.c
```diff
@@ -0,0 +1,425 @@
...
+     * invariant the semaphore enforces */
+    if (work->activated) {
+        return MENDER_OK;
+    }
+
+    /* Give semaphore used to protect the work function */
+    xSemaphoreGive(work->sem_handle);
+
+    /* Check the timer period */
+    if (work->params.period > 0) {
```
vpodzime: The Zephyr implementation doesn't allow the period to be `0` and I think we should do the same here.

src/platform/os/freertos/os.c
```diff
@@ -0,0 +1,425 @@
...
+            mender_log_error("Unable to start timer");
+            return MENDER_FAIL;
+        }
+
+        /* Execute the work now by enqueuing it */
+        mender_os_scheduler_timer_callback(work->timer_handle);
+    }
+
+    /* Indicate the work has been activated */
+    work->activated = true;
```
vpodzime, line 219: It's not really activated in case `.period == 0`, right?

vpodzime (commented):
I'd swap the second and third commit, the former is quite strange without the latter applied first.

src/platform/os/freertos/os.c
danielskinstad: It's the default in Joel's client. I don't have any other good reason for it.

src/platform/os/freertos/os.c
danielskinstad: I'll create two follow-up tickets in https://northerntech.atlassian.net/browse/MEN-8691.

src/platform/os/freertos/os.c
```diff
@@ -0,0 +1,425 @@
...
+    }
+
+    /* Copy work parameters.
+     * Note: backoff parameters are ignored; failed work simply retries at its
+     * regular period. Proper backoff and rate-limit handling is a follow-up. */
+    work_context->params.function = work_params->function;
+    work_context->params.period   = work_params->period;
+
+    if (NULL == (work_context->params.name = mender_utils_strdup(work_params->name))) {
+        mender_log_error("Unable to allocate memory");
```
michalkopczan, line 149: We could make the error more explicit. I know that the original issue is most probably lack of memory, but what actually matters from user perspective is that we could not assign a name to the work_context

michalkopczan (commented):
LGTM, but I'll leave the approval to Vratislav as he's more experienced with mender-mcu generally

src/platform/os/freertos/os.c
vpodzime: > we could not assign a name to the work_context

I'm not sure they care about that much detail. I'd say it's a responsibility of the function calling this one to log something explaining where exactly it happened. But generally I agree. We should have a macro for this specific `mender_log_error()` call that shows up in so many places. And it should mention the function name or `__FILE__`, `__LINE__` and such things, at least in debug builds. Let's make sure we follow up on this

michalkopczan: Yes, a function calling this one should print where it happened. But it would be nice to know what exactly happened in `mender_os_scheduler_work_create`. Line/name macro would be good.

For now, notice that for all other errors in this function we say exactly what failed: a semaphore could not be created, calloc failed, timer couldn't be created. But for this, we just say the reason for the fail, not what actually failed.

vpodzime: Yup, I understand your point. But let's fix/improve this across the codebase in a separate PR. The above line is just following the same pattern already present in tons of other places.

michalkopczan: Ah, OK. Didn't realize it is following a pattern, thank's for the explanation.

src/platform/os/freertos/os.c
danielskinstad: Right, good catch

vpodzime (approved):
Looks good to me now. Thanks!

PR #258, opened 2026-08-07
feat: Add support for a secondary/fallback server URL

vpodzime:
Always relying on a single Mender server host URL is dangerous
because if connections to this host URL start failing
persistently (i.e. if there are DNS/TLS/... issues), devices are
lost disconnected forever because there is no way to reach them.

To prevent such disasters, a secondary/fallback server host URL
should be set serving as a backup which will allow the device to
continue operations or at least to be updated with a new
configuration using a different primary server host URL.

Ticket: MEN-9977
Changelog: A secondary/fallback Mender server URL can now be specified
Signed-off-by: Vratislav Podzimek <vratislav.podzimek+auto-signed@northern.tech>


src/core/api.c
```diff
@@ -126,6 +129,20 @@ mender_api_init(mender_api_config_t *config) {
     return ret;
 }
 
+#ifndef CONFIG_MENDER_SECONDARY_SERVER_HOST_DISABLE
+static inline const char *
```
michalkopczan, line 137: I'd add a comment mentioning that the old host is returned. When I first read it I assumed it will return the new host and got surprised when I read `const char *old_host = swap_primary_secondary_host();`

src/core/client.c
```diff
@@ -322,6 +322,25 @@ mender_client_init(mender_client_config_t *config, mender_client_callbacks_t *ca
...
         goto END;
     }
+
+#ifndef CONFIG_MENDER_SECONDARY_SERVER_HOST_DISABLE
+    if ((NULL != config->secondary_host) && (strlen(config->secondary_host) > 0)) {
+        mender_client_config.secondary_host = config->secondary_host;
+    } else {
+        mender_client_config.secondary_host = CONFIG_MENDER_SECONDARY_SERVER_HOST;
+    }
+    if ((NULL == mender_client_config.secondary_host) || (0 == strlen(mender_client_config.secondary_host))) {
```
michalkopczan, line 332: Doesn't change the end behavior, but IMO we should first check if the secondary host is configured correctly and only then assign it. I.e. swap the order of `if ((NULL == mender_client_config.secondary_host) || (0 == strlen(mender_client_config.secondary_host)))` and `if ((NULL != config->secondary_host) && (strlen(config->secondary_host) > 0)`

src/core/client.c
```diff
@@ -622,6 +644,9 @@ mender_client_work_function(void) {
                     /* Try to release network so that it gets set up again next
                        time. */
                     mender_client_network_release();
+                    /* Also make sure we re-authenticate to the server, possibly
+                       using a fallback host/URL. */
+                    mender_api_drop_authentication_data();
```
michalkopczan, line 649: nice

target/zephyr/Kconfig
```diff
@@ -515,6 +531,12 @@ if MENDER_MCU_CLIENT
                 help
                     A secondary security tag that ROOT CA server credential will be referenced with, typically used to authenticate the server from where to download Mender Artifacts. See tls_credential_add.
 
+            config MENDER_NET_CA_CERTIFICATE_TAG_TERTIARY
+                int "Tertiary CA certificate tag for Artifacts"
```
michalkopczan: Shouldn't this say "Tertiary CA certificate tag for Server"? Or maybe do primary server + secondary server and tertiary for Artifacts?

michalkopczan (approved):
Besides some nitpicks and the comment about the Kconfig option descriptions, LGTM. I understand that the issues we discussed like not falling over immediately and instead after some time will be followed up separately?

vpodzime:
> I understand that the issues we discussed like not falling over immediately and instead after some time will be followed up separately?

In general, yes. But the non-instant fail-over happens with these changes:
- if the client successfully authenticates to the primary URL, 
- then some connection issues happen incrementing the network error counter, 
- only after a certain number of such errors counted the execution of the main worker function will drop the authentication data
- and new authentication will happen on a next request, potentially doing the fail-over.

An immediate fail-over can still happen here, though, if any regular authentication (the very first or after the token expires) with the primary URL fails. However, that's the same thing as the C++ client does, AFAICT. We need that bigger discussion for changes in this broader sense/behavior.

src/core/api.c
vpodzime: yup! I forgot to do that.

target/zephyr/Kconfig
vpodzime: yup, good catch!

vpodzime: I'd like to keep the secondary for artifacts as that's the default case with hosted Mender.

src/core/client.c
vpodzime: This way we have code that covers both the case using `config->secondary_host` and the case using `CONFIG_MENDER_SECONDARY_SERVER_HOST`.

michalkopczan: ah, I misread it. Clear.

danielskinstad (approved):
LGTM

PR #260, opened 2026-08-20
ci: use zephyr-4.2.x branch in mender-mcu-integration trigger

danielskinstad:
mender-mcu 1.0.x doesn't support zephyr 4.4 which is used in mender-mcu-integration main.


vpodzime (approved):
It just keeps giving! 😁

PR #264, opened 2026-08-25
fix(zephyr-image): Add a cleanup callback preventing memory leak in the zephyr-image UM

vpodzime:
In case of a download failure the state machine jumps directly into the `CLEANUP` state (there is no deployment to abort/fail yet). We need to make sure the flash handle used by the zephyr-image Update Module is released in such cases.

Ticket: None
Changelog: None

### DIFF
--- a/src/platform/update_modules/zephyr/image/update-module.c
+++ b/src/platform/update_modules/zephyr/image/update-module.c
@@ -189,6 +189,11 @@ static mender_err_t mender_zephyr_image_set_pending_image(mender_update_state_t 
 static mender_err_t mender_zephyr_image_abort_deployment(mender_update_state_t state, mender_update_state_data_t callback_data);
 
 /**
+ * @brief Cleanup callback
+ */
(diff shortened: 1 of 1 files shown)


src/platform/update_modules/zephyr/image/update-module.c
```diff
@@ -324,6 +330,15 @@ mender_zephyr_image_abort_deployment(MENDER_NDEBUG_UNUSED mender_update_state_t
     return MENDER_OK;
 }
 
+static mender_err_t
+mender_zephyr_image_cleanup(mender_update_state_t state, mender_update_state_data_t callback_data) {
+    assert(MENDER_UPDATE_STATE_CLEANUP == state);
+
+    /* Cleanup can be called without aborting a deployment in case of a download
+       failure. Let's make sure we don't leak the flash handle.  */
+    FREE_AND_NULL(mcu_boot_flash_handle);
```
danielskinstad, line 339: Missing a return

vpodzime: Who needs a return?! :sweat_smile:

danielskinstad (approved):
LGTM

PR #269, opened 2026-09-07
build: set mender-mcu version for esp-idf

danielskinstad:
Take the version from the component manager instead of hardcoding it. It gives the release version for registry components and the commit for git dependencies. Local path dependencies have neither, so use the exact tag or the commit from git directly.


michalkopczan (approved):

target/esp-idf/component.cmake
```diff
@@ -127,6 +127,19 @@ endif()
...
-# The component manager only fetches the staged files, so no git information is available
-target_compile_definitions(${COMPONENT_LIB} PUBLIC MENDER_CLIENT_VERSION="esp-idf-demo")
+
+idf_component_get_property(MENDER_MCU_VERSION ${COMPONENT_NAME} COMPONENT_VERSION)
+if(NOT MENDER_MCU_VERSION OR MENDER_MCU_VERSION STREQUAL "*")
+    # local path dependency: exact tag, else the commit
+    git_describe(MENDER_MCU_VERSION ${MENDER_MCU_ROOT} --exact-match)
+    if(MENDER_MCU_VERSION MATCHES "NOTFOUND")
+        set(MENDER_MCU_VERSION "unknown")
+        get_git_head_revision(_refspec MENDER_MCU_VERSION ${MENDER_MCU_ROOT})
```
vpodzime, line 137: Where do these extra `git` functions come from? Don't we need to include something extra or require some more recent version of CMake?

vpodzime (approved):
Looks good to me otherwise.

vpodzime:
@danielskinstad this needs a rebase :innocent:

target/esp-idf/component.cmake
danielskinstad: They come from esp-idf: https://github.com/espressif/esp-idf/blob/master/tools/cmake/third_party/GetGitRevisionDescription.cmake

